Issue #35, opened 2025-05-21
Mark routes with their response types for swagger docs


mferrera:
All possible exceptions that could be raised should be marked in the route decorator

PR #42, opened 2025-05-23
DOC: Add create session Swagger docs

mferrera:
Contributes to #35 

## Checklist

- [x] Tests added (not tests possible)
- [ ] Moved issue status on project board

### DIFF
--- a/src/fmu_settings_api/v1/responses.py
+++ b/src/fmu_settings_api/v1/responses.py
@@ -4,6 +4,63 @@ from typing import Any, Final, TypeAlias
 
 Responses: TypeAlias = dict[int | str, dict[str, Any]]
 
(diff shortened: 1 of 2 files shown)


src/fmu_settings_api/v1/responses.py
```diff
@@ -4,6 +4,52 @@
 
 Responses: TypeAlias = dict[int | str, dict[str, Any]]
 
+CreateSessionResponses: Final[Responses] = {
```
joargr, line 7: I see that when an attempt is made to create a session (`POST /api/v1/session`) and the API token is wrong, the response is `401 Unauthorized`. Should this response also be part of this list? It seems that the response code comes from the failed `AuthTokenDep` dependency, but from the outside it's a response from the session creation endpoint.

PR #70, opened 2025-06-06
ENH: Add SMDA interface and complete health route

### LINKED ISSUE #54: Add initial SMDA interface and routes
Add the initial SMDA query functionality when a user's SMDA subscription key is known.

mferrera:
Resolves #54 

## Checklist

### DIFF
--- a/src/fmu_settings_api/v1/routes/smda.py
+++ b/src/fmu_settings_api/v1/routes/smda.py
@@ -2,8 +2,10 @@
 
 from textwrap import dedent
 
-from fastapi import APIRouter
+from fastapi import APIRouter, HTTPException
 
+from fmu_settings_api.deps import SessionDep
+from fmu_settings_api.interfaces import SmdaAPI
 from fmu_settings_api.models import Ok
 from fmu_settings_api.v1.responses import GetSessionResponses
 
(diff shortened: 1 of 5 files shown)


src/fmu_settings_api/v1/routes/smda.py
```diff
@@ -31,6 +33,20 @@
         **GetSessionResponses,
     },
 )
-async def get_check() -> Ok:
+async def get_health(session: SessionDep) -> Ok:
     """Returns a simple 200 OK if able to query SMDA."""
-    return Ok()
+    # Handled on the route dependency, duplicated for typing
+    if session.access_tokens.smda_api is None:
+        raise HTTPException(status_code=401, detail="SMDA access token is not set")
```
joargr, line 40: Hmm, interesting. This is a situation where a call to the local API fails with 401 due to non-authorization in an external API. So the session to the local API is ok and doesn't result in a 401, but 401 will still be the result since it refers to the external API. Currently, getting the 401 result from this endpoint (and other SMDA endpoints) will cause the GUI to trigger a session creation. After 3 retries of this endpoint with a session creation for each retry, the user will see the error message and no more requests will be made.

Ideally the GUI should be more concise in triggering the session creation on 401 results, so that 401 results from SMDA won't get special handling. I will make an issue for this. Also, perhaps the GUI should be able to treat an SMDA 401 in such a way as to guide the user to enter the SMDA subscription key.

PR #80, opened 2025-06-19
FIX: Return external API code as own

### LINKED ISSUE #77: Return external API http codes as own
Add a secondary field in the exception to indicate if the error is external or internal

mferrera:
Resolves #77

## Checklist


joargr (approved):
Interesting to set an HTTP header to convey information that this is a response code coming from an external API and not the local API. I think this can be picked up by the error handling in the GUI (the Axios interceptor), so that the correct handling can be done: Don't create a session on 401 if it comes from an external API.

PR #99, opened 2025-08-28
ENH: Add endpoint to load global_config

### LINKED ISSUE #89: API: Import masterdata from existing global configuration
Models already using Sumo will have their masterdata set somewhere. We should allow importing this from the global configuration.

To be able to do this we should add a new endpoint `/project/global_config` (POST) to `fmu-settings-api` to import masterdata from existing global configuration into the project config. 

The endpoint should take a path as an optional input parameter. If the path is set it should look for the global config file at the provided path. If the path is not set, it should search for the global config at the default location for the global config. 

The endpoint should try to import the global config from the default location or the provided path. If the global config is not found, return 404.: Not found

If a global config exists, the global config should be validated: 
* If validation OK, update the project config with the global config masterdata and return "OK, project exists" to GUI (dont return config).
* If validation failed, return an error message and error code 500 to the GUI, stating that global config is not valid (dont return config).

The endpoint should use the GlobalConfigManager in `fmu-settings` to search for, read and validate the global config from disk. This will be implemented through https://github.com/equinor/fmu-config/issues/97

If a valid config is found, the ConfigManager in `fmu-settings` can be used to update the masterdata.

slangeveld:
Resolves #89 

Add endpoint to import global config from `global_variables.yml` into the project config. In this first iteration only the masterdata is loaded into the project masterdata.

The endpoint accepts and optional path as input. If the path is not set, the endpoint will look for the `global_variables.yml` file at the default location `<project_root>/fmuconfig/output/global_variables.yml`. If a path is provided as a parameter in the request, the endpoint will look for the `global_variables.yml` file at the provided path.

* If the `global_variables.yml` file is not found, a 404 status code is returned with a descriptional message.
* If the `global_variables.yml` file does not validate agains the Pydantic model GlobalConfiguration, a 500 status code is returned with the validation error message. 
* If the user does not have a project session, a 401 status code is retuned (from the injected dependency ProjectSessionDep).
 

## Checklist

### DIFF
--- a/src/fmu_settings_api/v1/routes/project.py
+++ b/src/fmu_settings_api/v1/routes/project.py
@@ -5,15 +5,19 @@ from textwrap import dedent
 from typing import Final
 
 from fastapi import APIRouter, HTTPException, Response
+from fmu.config import utilities
+from fmu.datamodels.fmu_results import global_configuration
 from fmu.datamodels.fmu_results.fields import Smda
 from fmu.settings import find_nearest_fmu_directory, get_fmu_directory
 from fmu.settings._init import init_fmu_directory
+from pydantic import ValidationError
 
 from fmu_settings_api.deps import (
     ProjectSessionDep,
     SessionDep,
 )
 from fmu_settings_api.models import FMUDirPath, FMUProject, Message
+from fmu_settings_api.models.common import Ok
 from fmu_settings_api.session import (
     ProjectSession,
     SessionNotFoundError,
@@ -26,6 +30,7 @@ from fmu_settings_api.v1.responses import (
     inline_add_response,
 )
 
+GLOBAL_CONFIG_DEFAULT_PATH: Final[Path] = Path("fmuconfig/output/global_variables.yml")
 router = APIRouter(prefix="/project", tags=["project"])
 
 ProjectResponses: Final[Responses] = {
@@ -223,6 +228,57 @@ async def init_project(
         raise HTTPException(
             status_code=409, detail=f".fmu already exists at {path}"
         ) from e
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e)) from e
+
+
+@router.post(
+    "/global_config",
+    response_model=Ok,
+    summary=("Loads the global config into the project masterdata."),
+    description=dedent(
+        """
+        Loads the global config into the project masterdata. If the
+        global config does not validate, or is not found, a failed status code is
+        returned. The global config is searched for at the default project path,
+        or at the path provided as a parameter in the request.
+       """
+    ),
+    responses={
+        **GetSessionResponses,
+        **ProjectResponses,
+        **ProjectExistsResponses,
+    },
+)
+async def load_global_config(
+    project_session: ProjectSessionDep, global_config_path: Path | None = None
+) -> Ok:
+    """Loads the global config into the project masterdata if found and valid."""
+    try:
+        fmu_dir = project_session.project_fmu_directory
+        mode = "provided"
+        if global_config_path is None:
+            project_path = fmu_dir.get_file_path(
+                fmu_dir.config.relative_path
+            ).parent.parent
+            global_config_path = project_path / GLOBAL_CONFIG_DEFAULT_PATH
+            mode = "default"
+
+        if not global_config_path.exists():
+            raise FileExistsError(
+                f"Could not find the global_variables.yml file at the {mode} path."
+            )
+
+        global_config_dict = utilities.yaml_load(global_config_path)
+        global_config = global_configuration.GlobalConfiguration.model_validate(
+            global_config_dict
+        )
+        fmu_dir.set_config_value("masterdata", global_config.masterdata.model_dump())
+        return Ok()
+    except FileExistsError as e:
+        raise HTTPException(status_code=404, detail=str(e)) from e
+    except ValidationError as e:
+        raise HTTPException(status_code=500, detail=str(e)) from e
     except Exception as e:
         raise HTTPException(status_code=500, detail=str(e)) from e
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -177,3 +177,54 @@ def smda_masterdata() -> dict[str, Any]:
             {"identifier": "OseFax", "uuid": "15ce3b84-766f-4c93-9050-b154861f9100"}
         ],
     }
+
+
+@pytest.fixture
+def global_variables_mock() -> dict[str, Any]:
+    """Returns an example of the global_variables.yml file with smda masterdata."""
+    return {
+        "masterdata": {
+            "smda": {
+                "stratigraphic_column": {
+                    "identifier": "DROGON_2020",
+                    "uuid": "15ce3b84-766f-4c93-9050-b154861f9100",
+                },
+                "coordinate_system": {
+                    "identifier": "ST_WGS84_UTM37N_P32637",
+                    "uuid": "15ce3b84-766f-4c93-9050-b154861f9100",
+                },
+                "country": [
+                    {
+                        "identifier": "Norway",
+                        "uuid": "15ce3b84-766f-4c93-9050-b154861f9100",
+                    }
+                ],
+                "discovery": [
+                    {
+                        "short_identifier": "SomeDiscovery",
+                        "uuid": "15ce3b84-766f-4c93-9050-b154861f9100",
+                    }
+                ],
+                "field": [
+                    {
+                        "identifier": "OseFax",
+                        "uuid": "15ce3b84-766f-4c93-9050-b154861f9100",
+                    }
+                ],
+            }
+        },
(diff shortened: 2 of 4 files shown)


tests/conftest.py
```diff
@@ -177,3 +177,54 @@ def smda_masterdata() -> dict[str, Any]:
             {"identifier": "OseFax", "uuid": "15ce3b84-766f-4c93-9050-b154861f9100"}
         ],
     }
+
+
+@pytest.fixture
+def global_variables_mock() -> dict[str, Any]:
+    """Returns an example of the global_variables.yml file with smda masterdata."""
```
joargr, line 184: For consistency the spelling here should be "SMDA", with capital letters as it is pure text and not part of code.

tests/test_v1/test_project.py
```diff
@@ -572,3 +574,165 @@ def test_patch_masterdata_no_directory(
...
+    )
+    assert (
+        fmu_project.config.masterdata.smda.country[0].identifier
+        == expected_smda_country
+    )
+
+
+def test_load_global_config_file_not_found(
+    client_with_project_session: TestClient, session_tmp_path: Path
+) -> None:
```
joargr, line 690: There should perhaps be separate tests for not-found for default path and for provided path, in the same way that the successful import has separate tests for these two modes.

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -227,6 +232,57 @@ async def init_project(
...
+    "/global_config",
+    response_model=Ok,
+    summary=("Loads the global config into the project masterdata."),
+    description=dedent(
+        """
+        Loads the global config into the project masterdata. If the
+        global config does not validate, or is not found, a failed status code is
+        returned. The global config is searched for at the default project path,
+        or at the path provided as a parameter in the request.
+       """
```
joargr, line 245: Should the given global config path be a relative path, relative to the project root, or will it be allowed to be an absolute path? Is there a use case where a user needs to import a global config file from a different project? And if so, will it be allowed to give this absolute path or does the user then need to copy the global config file to somewhere within the project first?

I think it's best to just allow relative paths, and to make sure that the supplied relative path does not resolve to somewhere outside of the project. In `fmu-settings-gui:src/fmu_settings_gui/__main__.py`, in the function `serve_spa_catchall`, there is code that checks and sanitises a user provided path (using functions that satisfies the GitHub CodeQL code check). Something similar should be done in the import function.

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -227,6 +232,57 @@ async def init_project(
...
+    ),
+    responses={
+        **GetSessionResponses,
+        **ProjectResponses,
+        **ProjectExistsResponses,
+    },
+)
+async def load_global_config(
+    project_session: ProjectSessionDep, global_config_path: Path | None = None
+) -> Ok:
```
joargr, line 255: The parameters - the `global_config_path` parameter - should be received in the request body, and not as query parameters. In FastAPI this will happen when the parameter is defined using Pydantic. The implementation of `POST /project` should be quite similar to how it can be done for the import endpoint. Also, I think the parameter can be called `path`, as the "global config" part of the name is implied from the endpoint description.

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -227,6 +232,57 @@ async def init_project(
...
+        if global_config_path is None:
+            project_path = fmu_dir.get_file_path(
+                fmu_dir.config.relative_path
+            ).parent.parent
+            global_config_path = project_path / GLOBAL_CONFIG_DEFAULT_PATH
+            mode = "default"
+
+        if not global_config_path.exists():
+            raise FileExistsError(
+                f"Could not find the global_variables.yml file at the {mode} path."
```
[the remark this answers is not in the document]
slangeveld: Good point

src/fmu_settings_api/v1/routes/project.py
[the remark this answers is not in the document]
slangeveld: This will be solved through https://github.com/equinor/fmu-settings-api/issues/100 in a separate endpoint

src/fmu_settings_api/v1/routes/project.py
slangeveld: My thought was that it should allow for any absolute path, but I see your point of only allowing relative paths to avoid imports from outside the project. What would a typical usecase for a user be here @tnatt?

tnatt: I don't see the need for importing outside the existing project no, so we can add that requirement 👍

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -227,6 +232,57 @@ async def init_project(
...
+    project_session: ProjectSessionDep, global_config_path: Path | None = None
+) -> Ok:
+    """Loads the global config into the project masterdata if found and valid."""
+    try:
+        fmu_dir = project_session.project_fmu_directory
+        mode = "provided"
+        if global_config_path is None:
+            project_path = fmu_dir.get_file_path(
+                fmu_dir.config.relative_path
+            ).parent.parent
```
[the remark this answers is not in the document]
tnatt: You should be able to find the `project_path` through the property `fmu_dir.path` 🙂  https://github.com/equinor/fmu-settings/blob/6f520dcfb53856b034fded66fde2bb3bce02b7d8/src/fmu/settings/_fmu_dir.py#L53-L55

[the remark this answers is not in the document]
slangeveld: Ahh, of course :see_no_evil: Thansk

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -227,6 +232,57 @@ async def init_project(
...
+            raise FileExistsError(
+                f"Could not find the global_variables.yml file at the {mode} path."
+            )
+
+        global_config_dict = utilities.yaml_load(global_config_path)
+        global_config = global_configuration.GlobalConfiguration.model_validate(
+            global_config_dict
+        )
+        fmu_dir.set_config_value("masterdata", global_config.masterdata.model_dump())
+        return Ok()
```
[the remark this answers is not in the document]
slangeveld: Yeah, was a bit uncertain about this on. Mostly picked OK since that was what was asked for from the GUI side. But maybe it makes more sense to `Message` for this POST endpoint and use OK for the other `/check_global_config` https://github.com/equinor/fmu-settings-api/issues/100 endpoint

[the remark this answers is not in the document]
joargr: I think we can continue the pattern of returning a `Message`, which includes a text. Then the GUI can decide based on the user interface how and if to display the message text, or whether in some cases it's preferred to display a text coming from the GUI itself. But it makes sense to have a message also for successful requests, in the same was that failed requests have a message. For many endpoints we transform the API response into a GUI object with a status (successful/failed), a text and data. Having a `Message` object in the response also for a successful request makes such a transformation natural.

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -235,50 +279,56 @@ async def init_project(
...
     description=dedent(
         """
-        Loads the global config into the project masterdata. If the
-        global config does not validate, or is not found, a failed status code is
-        returned. The global config is searched for at the default project path,
-        or at the path provided as a parameter in the request.
+        Loads the global config into the project masterdata. If the global config does
+        not validate, or is not found, a failed status code is returned. The endpoint
+        takes an optional parameter, `path` as input: This should be given as a relative
+        path, relative to the project root. If provided, the global config is searched
```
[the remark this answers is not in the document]
slangeveld: Ok, will leave it as it then

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -227,6 +276,63 @@ async def init_project(
...
+        raise HTTPException(
+            status_code=409,
+            detail="A config file with masterdata already exists in "
+            f".fmu at {fmu_dir.config.path}.",
+        ) from e
+    except ValidationError as e:
+        raise HTTPException(
+            status_code=500,
+            detail=f"The global config file is not valid at path {global_config_path}.",
+        ) from e
```
joargr, line 331: When the referenced file has validation errors I think it makes more sense to use the HTTP status code `422 Unprocessable Content`. This is a client error, whereas `500 Internal Server Error` is a server error. A client error means that the error is caused by something supplied by the client, in this case a reference to a global config file which turns out to be of invalid format. The 500 server error should be used for unexpected errors, typically as a last resort when handling exceptions. In the GUI responses with 500 errors are generally handled by display a toast. If the global config file to be imported turns out to have an invalid format, the error message should be displayed as part of the import form, in the same way that specifying a non-existing file should display the 404 error message as part of the import form.

slangeveld: Fair point. Will update to 422

PR #114, opened 2025-09-17
ENH: Drop non-existing project from recent_projects

### LINKED ISSUE #102: Remove non-existing recent_project_directories
When a trying to get a project path that is non-existing it should be removed from the recent_project_directories list if present there.

Users will have the ability to select from the recent project directories in the GUI. And it will be a common scenario for some paths to no longer exist, as user copies of projects are frequently deleted.
The GUI can then respond and refetch user data.

tnatt:
Resolves #102 

This PR makes a change to the `POST /project` endpoint so that when a non-existing path is input it is removed from the users `recent_project_directories` if it is present there.

## Checklist

- [ ] Appropriate [commit prefix](https://upgraded-funicular-eywe4gy.pages.github.io/developing/#commit-prefixes) and precise commit message used
- [ ] Latest main rebased/merged into branch
- [ ] Moved issue status on project board
- [ ] Checked the boxes in this checklist ✅


tests/test_services/test_user_service.py
```diff
@@ -69,3 +72,32 @@ def test_add_to_user_recent_projects_removes_oldest_when_full(
...
+    user_dir.set_config_value("recent_project_directories", [non_existing_user_copy])
+
+    remove_from_recent_projects(non_existing_user_copy, user_dir)
+    assert user_dir.get_config_value("recent_project_directories") == []
+
+
+def test_remove_from_user_recent_projects_does_not_remove_existing(
+    tmp_path_mocked_home: Path,
+) -> None:
+    """Tests removing an path not present does not modify recent projects."""
```
slangeveld, line 92: Minor typo here a/an

slangeveld (approved):
Looks good 👍

Issue #118, opened 2025-09-26
Using deprecated 422 code constant


mferrera:
```sh
tests/test_v1/test_smda.py::test_post_field_with_no_json_fails
  /opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/asyncio/events.py:88: DeprecationWarning: 'HTTP_422_UNPROCESSABLE_ENTITY' is deprecated. Use 'HTTP_422_UNPROCESSABLE_CONTENT' instead.
```

There is at least one other place as well.

PR #125, opened 2025-10-02
ENH: Add PATCH model endpoint to save model to .fmu config

### LINKED ISSUE #123: Add endpoint for setting `model`
Currently we only have an endpoint for saving `masterdata` to the project config `PATCH project/masterdata`
We also need an endpoint for setting the `model` block in the project `.fmu` config.

It should be very similar to the `patch_masterdata` function except we want to accept a `model` argument of type [Model](https://github.com/equinor/fmu-datamodels/blob/6a2409813388277dd78e6864106375d8bb9a8a6d/src/fmu/datamodels/fmu_results/fields.py#L236) 
https://github.com/equinor/fmu-settings-api/blob/16c87db30cfe20cb1601a4f9d057df914aebc399/src/fmu_settings_api/v1/routes/project.py#L435-L449

GibranAlfa:
Resolves #123 #118 

- Added PATCH model endpoint to save model to .fmu config
- Added unit tests for that endpoint
- Update HTTP_422_UNPROCESSABLE_ENTITY to HTTP_422_UNPROCESSABLE_CONTENT to address warnings

## Checklist

### DIFF
--- a/tests/test_v1/test_smda.py
+++ b/tests/test_v1/test_smda.py
@@ -267,7 +267,9 @@ async def test_post_field_with_no_json_fails(
     """Tests that posting without json causes Pydantic validation errors."""
     response = client_with_smda_session.post(f"{ROUTE}/field")
 
-    assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY, response.json()
+    assert response.status_code == status.HTTP_422_UNPROCESSABLE_CONTENT, (
+        response.json()
+    )
     assert response.json()["detail"] == [
         {
             "input": None,


src/fmu_settings_api/v1/routes/project.py
```diff
@@ -449,6 +449,38 @@ async def patch_masterdata(
...
 
+@router.patch(
+    "/model",
+    response_model=Message,
+    summary="Saves model information to the project .fmu directory",
+    description=dedent(
+        """
+        Saves model information to the project .fmu directory.
+        If existing model information is present, it will be updated with the new
+        model data.
```
joargr, line 460: As the new model data overwrites any existing model data, it should be more precise to say "it will be replaced by the new model data". Also, perhaps the term to use should be "model data", used consistently.

Issue #47, opened 2025-05-26
Posting a new project path should trigger clean-up logic if a project is already open


mferrera:
Posting a new project should change the project .fmu session to the new project, but ensures that if an existing project was opened, it is closed and cleaned up if necessary.

This basically relates to https://github.com/equinor/fmu-settings-api/issues/13 where a lock should be released, any opened files should be closed, etc. These thinks should happen when a particular project session is closed.

PR #136, opened 2025-10-06
ENH: Acquire .fmu lock when FMU Directory is opened

### LINKED ISSUE #13: Acquire .fmu lock when FMU Directory is opened
When adding a project .fmu directory to a session, it should acquire the lock on it. The lock is present on an `project_fmu_directory` instance and can be called with `project_fmu_directory._lock.acquire()`.

This lock _must_ be released under failures circumstances, and must be released when a project .fmu directory is closed/switched. This means there should be a `finally:` block somewhere that checks if a project .fmu directory is still open and if so, called `project_fmu_directory._lock.release()` on it.

It should not fail if a lock is already acquired, i.e. it should be able to inform the GUI that a lock cannot be acquired. In this circumstance data can still be read, but it _cannot_ be written.

That is, if a lock is held by someone else (not the current session), the API/GUI cannot write/save files to disk, and must be informed of this.

GibranAlfa:
Resolves #13 #47 

- Lock acquisition when FMU directories are opened
- Lock release when projects are closed or sessions destroyed
- GUI awareness through is_read_only field
- Add write protection if in read only mode

The flow:
1. Open project
2. Try to acquire lock
3. If locked, the session remains in read only mode
4. If not locked, the lock is acquired
5. Routes that try to writes check whether we are in read-only mode or not
6. If in read-only mode, block write operation

## Checklist

### DIFF
--- a/src/fmu_settings_api/v1/routes/project.py
+++ b/src/fmu_settings_api/v1/routes/project.py
@@ -158,7 +158,6 @@ GlobalConfigResponses: Final[Responses] = {
     responses={
         **GetSessionResponses,
         **ProjectResponses,
-        **LockConflictResponses,
     },
 )
 async def get_project(session: SessionDep) -> FMUProject:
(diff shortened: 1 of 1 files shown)


slangeveld (commented):
Nice work. A few comments to consider

PR #143, opened 2025-10-14
ENH: Refresh fmu dir lock when session is refreshed

### LINKED ISSUE #135: Refresh fmu directory lock when session is refreshed
Sessions are refreshed by the GUI from time to time so they aren't expired as stale.

When this happens, we should also refresh the .fmu directory lock, if one is held, to ensure that a lock continues to be held with the session. But, only so long as a project .fmu is held open by the session.

The lock manager in the .fmu directory has a `refresh()` method to do this.

GibranAlfa:
Resolves #135 

- Add refresh lock in `get_session` in `SessionManager`
- Add `last_lock_refresh_error` to track error when refreshing the lock and provide it in `lock_status` endpoint

## Checklist

### DIFF
--- a/src/fmu_settings_api/session.py
+++ b/src/fmu_settings_api/session.py
@@ -1,6 +1,6 @@
 """Functionality for managing sessions."""
 
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, field
 from datetime import UTC, datetime, timedelta
 from typing import Self
 from uuid import uuid4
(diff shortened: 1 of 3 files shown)


src/fmu_settings_api/session.py
```diff
@@ -198,12 +211,14 @@ async def add_fmu_project_to_session(
         project_session.project_fmu_directory = project_fmu_directory
         project_session.last_lock_acquire_error = last_lock_acquire_error
         project_session.last_lock_release_error = last_lock_release_error
+        project_session.last_lock_refresh_error = None
```
mferrera: Should these grow into their own key, perhaps as a dataclass?

```python
project_session.lock_errors: LockErrors = LockErrors(acquire="..", release="..", refresh="..")
```

(or something)

PR #160, opened 2025-10-24
ENH: Allow sliding session expiration

### LINKED ISSUE #156: Increase session expiration time
https://github.com/equinor/fmu-settings-api/blob/1dbaa2629f1284779a1d6ff487e61580b975169a/src/fmu_settings_api/session.py#L156-L157

The last_accessed is updated, but the expiration is not. The cookie may need to be updated here also.

mferrera:
In most webapps it is an anti-pattern/security risk to have sliding session expirations based upon requests. In our use case, however, with shared same-site resources, it makes sense to. This ensures new sessions do not need to be recreated as long as the user is active and allows the GUI to think less about session expiration itself.

Resolves #156

## Checklist

- [ ] Appropriate [commit prefix](https://upgraded-funicular-eywe4gy.pages.github.io/developing/#commit-prefixes) and precise commit message used

### DIFF
--- a/src/fmu_settings_api/session.py
+++ b/src/fmu_settings_api/session.py
@@ -132,7 +132,9 @@ class SessionManager:
 
         return session_id
 
-    async def get_session(self: Self, session_id: str) -> Session | ProjectSession:
+    async def get_session(
+        self: Self, session_id: str, extend_expiration: bool = True
+    ) -> Session | ProjectSession:
         """Get the session data for a session id.
 
         Params:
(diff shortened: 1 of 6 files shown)


src/fmu_settings_api/v1/main.py
```diff
@@ -10,7 +10,7 @@
 from .routes.smda import main as smda
 
 api_v1_router = APIRouter()
-api_v1_router.include_router(project.router, dependencies=[Depends(get_session)])
+api_v1_router.include_router(project.router)
```
mferrera: This is a not-small change. But, every current `/project` route requires a reference to the session, so also nothing has changed. It just means new routes need to ensure they include the dependency in manually, which they need to do for the reference anyway.

src/fmu_settings_api/session.py
```diff
@@ -155,6 +157,10 @@ async def get_session(self: Self, session_id: str) -> Session | ProjectSession:
 
         session.last_accessed = now
 
+        if extend_expiration:
+            expiration_duration = timedelta(seconds=settings.SESSION_EXPIRE_SECONDS)
+            session.expires_at = now + expiration_duration
+
```
mferrera: Every renewal increases the session by the timeout length. This means the user will always have 20 minutes of inactivity needed before the API expires the session id, and the GUI doesn't need to track that.

PR #165, opened 2025-10-28
ENH: Add get session route

### LINKED ISSUE #158: Add 'get session' route to query session information
This allows the GUI an easier time of keeping in sync with the session state on the API.

It should return the session model to the API. For this reason, the session models may as well be converted to Pydantic models.

Some items should be removed:

- Access Tokens
- *FMUDirectory

This might be achieved by setting a serialization rules (exclude), or by a better splitting of models. (Maybe the session models should remain dataclasses, and response models are created from them, f.ex)

GibranAlfa:
Resolves #158 

- Add get session route to query session information
- Add `SessionResponse` model based on in-memory session state. Removed `user_fmu_directory`(is info about user fmu dir needed?) and `access_tokens`. Include optional info about the project and project lock errors

## Checklist

### DIFF
--- a/tests/test_v1/test_session.py
+++ b/tests/test_v1/test_session.py
@@ -295,6 +295,26 @@ async def test_get_session_returns_sanitised_payload(
     assert "access_tokens" not in payload
 
 
+async def test_get_session_does_not_extend_expiration(
+    client_with_session: TestClient,
+    session_manager: SessionManager,
+) -> None:
+    """Tests that GET /session should not extend session expiration."""
+    session_id = client_with_session.cookies.get(settings.SESSION_COOKIE_KEY)
+    assert session_id is not None
+
+    session = await session_manager.get_session(session_id)
+    original_expires_at = session.expires_at
+
+    response = client_with_session.get(ROUTE)
+    assert response.status_code == status.HTTP_200_OK
+
+    refreshed_session = await session_manager.get_session(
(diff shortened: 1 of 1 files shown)


src/fmu_settings_api/v1/routes/session.py
GibranAlfa: dep is already checked here

src/fmu_settings_api/session.py
```diff
@@ -304,3 +306,28 @@ async def add_access_token_to_session(session_id: str, token: AccessToken) -> No
...
+            is_read_only=not fmu_dir._lock.is_acquired(),
+        )
+        project_lock_errors = session.lock_errors.model_dump()
+
+    return SessionResponse(
+        id=session.id,
+        created_at=session.created_at,
+        expires_at=session.expires_at,
+        last_accessed=session.last_accessed,
+        project=project_response,
```
tnatt: I don't think we want to load and include the entire project config for this response, it would be sufficient with the `session.project_fmu_directory`.

tnatt (commented):
One comment from my side to consider 🙂 
The most important with this endpoint is for the GUI to know the timestamps i.e. when a session is created, last accessed and most importantly when it expires, getting the entire project here should not be necessary.

mferrera (approved):
🚀

PR #169, opened 2025-10-29
MAINT: Refactor SMDA route to use `SMDAService`

### LINKED ISSUE #145: Refactor smda routes to fully use an SmdaService
Currently a fair bit of business logic sits in the routes call.

https://github.com/equinor/fmu-settings-api/blob/de12ef13760d06ed03c7dbc8cce415d642780f09/src/fmu_settings_api/v1/routes/smda/main.py#L179-L292

This should be refactored out into `services/smda.py`

It may also be prudent to capture it in a class, like `SmdaService`. Then we can reduce the routes to a pattern that follows something like

```python
def get_smda_service() -> SmdaService:
    return SmdaService()  # lru_cache it?

# SmdaService also contains an instance of the SmdaInterface
SmdaServiceDep = Annotated[SmdaService, Depends(get_smda_service)]
```

```python
async def post_masterdata(
    smda_fields: list[SmdaField],
    smda_service: SmdaServiceDep,
) -> SmdaMasterdataResult:
    """Queries SMDA masterdata for .fmu project configuration."""
    return smda_service.get_masterdata(smda_fields)  # Handles all the error condition raising too
```

Separating the business logic away from the route will make it easier to refactor, or add other options, such as querying SMDA through Sumo core.

GibranAlfa:
Resolves #145 

Refactor SMDA route to use `SMDAService`.

## Checklist


src/fmu_settings_api/services/smda.py
```diff
@@ -1,94 +1,186 @@
...
+    SmdaField,
+    SmdaFieldSearchResult,
+    SmdaMasterdataResult,
+)
+
+
+class SmdaService:
+    """Service for querying SMDA API and handling business logic."""
+
+    def __init__(self, smda_api: SmdaAPI):
```
mferrera:
```suggestion
    def __init__(self, smda_api: SmdaAPI) -> None:
```

Nit

mferrera (approved):
Nice ✅

PR #170, opened 2025-10-30
MAINT: Refactor project routes to use a `ProjectService`

### LINKED ISSUE #146: Refactor `project/` routes to use a `ProjectService`
See #145 for inspiration.

It should be a dependency injection that injects `services.project_service.py :: ProjectService` to abstract the business logic away from the routes. All routes should then just be a simple method call to the passed `ProjectService`.

GibranAlfa:
Resolves #146 

`ProjectService` will handle business logic related to FMU project only. `SessionService` will handle session-level logic and lock management. I think that's a good separation

## Checklist

### DIFF
--- a/src/fmu_settings_api/deps/project.py
+++ b/src/fmu_settings_api/deps/project.py
@@ -6,7 +6,7 @@ from fastapi import Depends
 
 from fmu_settings_api.services.project import ProjectService
 
-from .session import ProjectSessionDep, ProjectSessionNoExtendDep
+from .session import ProjectSessionDep
 
 
 async def get_project_service(
@@ -16,14 +16,4 @@ async def get_project_service(
     return ProjectService(project_session.project_fmu_directory)
 
 
-async def get_project_service_no_extend(
-    project_session: ProjectSessionNoExtendDep,
-) -> ProjectService:
-    """Returns a ProjectService instance without extending session expiration."""
-    return ProjectService(project_session.project_fmu_directory)
-
-
 ProjectServiceDep = Annotated[ProjectService, Depends(get_project_service)]
-ProjectServiceNoExtendDep = Annotated[
-    ProjectService, Depends(get_project_service_no_extend)
-]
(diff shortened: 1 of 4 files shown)


src/fmu_settings_api/v1/routes/project.py
```diff
@@ -505,47 +479,18 @@ async def post_lock_acquire(project_session: ProjectSessionDep) -> Message:
         **ProjectResponses,
     },
 )
-async def get_lock_status(project_session: ProjectSessionNoExtendDep) -> LockStatus:
+async def get_lock_status(
+    project_session: ProjectSessionNoExtendDep,
```
GibranAlfa: this will be cleaner later when we have `SessionService`

src/fmu_settings_api/services/project.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def validate_global_config(self) -> None:
+        """Check if a valid global config exists at the default location."""
+        project_root = self._fmu_dir.path.parent
+        existing_config = find_global_config(project_root)
+
+        if existing_config is None:
+            raise FileNotFoundError("No valid global config file found in the project.")
+
+    def load_global_config(self, path: GlobalConfigPath | None = None) -> str:
```
slangeveld: This method name should probably also say a bit more on what it actually does. From reading the code now in the route `post_global_config` it looks a bit like it is loading the global config into the message variable, but there is actually also a write here under the hood.

src/fmu_settings_api/deps/project.py
slangeveld: Maybe add some simple tests for this new project dependency/deps? Would be good to at least test the that session expiration is set/not set.

slangeveld (approved):
Nice! Makes the code nice and clean! 

A few comments on some of the method names: It is a good thing to abstract away the the business code into a service layer, but it is important that we do it in such a way that it is still easy to understand what is happening on the route layer, without having to read the code in the service layer.

PR #171, opened 2025-10-31
MAINT: Refactor session route to use a `SessionService`

### LINKED ISSUE #147: Refactor `session/` routes to use a `SessionService`
See #145 for inspiration.

It should be a dependency injection that injects `services.session_service.py :: SessionService` to abstract the business logic away from the routes. All routes should then just be a simple method call to the passed `SessionService`.

The dependency function can receive the transient dependencies the and create a `SessionService` with them.

GibranAlfa:
Resolves #147 

Refactor session route to use a `SessionService`. Added tests to the new deps

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/project.py
+++ b/src/fmu_settings_api/services/project.py
@@ -1,5 +1,7 @@
 """Service for managing FMU project operations and business logic."""
 
+from pathlib import Path
+
 from fmu.datamodels.fmu_results.fields import Access, Model, Smda
 from fmu.settings import ProjectFMUDirectory
 from fmu.settings._global_config import find_global_config
@@ -57,17 +59,17 @@ class ProjectService:
             "masterdata", global_config.masterdata.model_dump()
         )
 
-    def update_masterdata(self, smda_masterdata: Smda) -> tuple[bool, str]:
+    def update_masterdata(self, smda_masterdata: Smda) -> tuple[bool, Path]:
         """Save SMDA masterdata to the project FMU directory."""
         self._fmu_dir.set_config_value("masterdata.smda", smda_masterdata.model_dump())
-        return True, f"Saved SMDA masterdata to {self._fmu_dir.path}"
+        return True, self._fmu_dir.path
 
-    def update_model(self, model: Model) -> tuple[bool, str]:
+    def update_model(self, model: Model) -> tuple[bool, Path]:
         """Save model data to the project FMU directory."""
         self._fmu_dir.set_config_value("model", model.model_dump())
-        return True, f"Saved model data to {self._fmu_dir.path}"
+        return True, self._fmu_dir.path
 
-    def update_access(self, access: Access) -> tuple[bool, str]:
+    def update_access(self, access: Access) -> tuple[bool, Path]:
         """Save access data to the project FMU directory."""
         self._fmu_dir.set_config_value("access", access.model_dump())
-        return True, f"Saved access data to {self._fmu_dir.path}"
+        return True, self._fmu_dir.path
--- a/src/fmu_settings_api/services/session.py
+++ b/src/fmu_settings_api/services/session.py
@@ -40,10 +40,10 @@ class SessionService:
             last_accessed=self._session.last_accessed,
         )
 
-    async def add_access_token(self, access_token: AccessToken) -> tuple[bool, str]:
+    async def add_access_token(self, access_token: AccessToken) -> str:
         """Add a known access token to the session."""
         await add_token_to_session_manager(self._session.id, access_token)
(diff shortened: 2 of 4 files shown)


src/fmu_settings_api/services/session.py
```diff
@@ -0,0 +1,142 @@
...
+        return SessionResponse(
+            id=self._session.id,
+            created_at=self._session.created_at,
+            expires_at=self._session.expires_at,
+            last_accessed=self._session.last_accessed,
+        )
+
+    async def add_access_token(self, access_token: AccessToken) -> str:
+        """Add a known access token to the session."""
+        await add_token_to_session_manager(self._session.id, access_token)
```
mferrera: And here too with returning an indicator other than a string. It might be nice to get a second opinion on this as well.

mferrera (commented):
This looks good. Some comments about functions that change state but return messages rather than the fail/success of the state. I think @slangeveld should chip in too,

src/fmu_settings_api/services/session.py
```diff
@@ -0,0 +1,142 @@
...
+        return fmu_dir
+
+    async def close_project(self) -> str:
+        """Remove (close) a project .fmu directory from the session."""
+        project_session = cast("ProjectSession", self._session)
+        project_path = project_session.project_fmu_directory.path
+        await remove_fmu_project_from_session(self._session.id)
+        return f"FMU directory {project_path} closed successfully"
+
+    async def acquire_project_lock(self) -> str:
```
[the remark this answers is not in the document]
GibranAlfa: agree 🫡

GibranAlfa:
> > This looks good. Some comments about functions that change state but return messages rather than the fail/success of the state. I think @slangeveld should chip in too,
> 
> Sorry for being late to the discussion here, as I see already some changes have been made. Regarding the service layer methods that return a message:
> 
> For me the whole `Message` concept belongs to the route/api layer and is not something the service layer should know about. The service layer should take care of handling the business logic and only that. The api layer can then use it's services, but the communication with the GUI should happen only from the API layer. By constructing the message to return to the user in the service layer we kindoff add a dependency between the gui and the service layer. So would probably just return a bool from the service methods and construct the whole message in the api layer, based on the result from the service method.

will update the `ProjectService` as well to return bool instead of str

src/fmu_settings_api/services/project.py
```diff
@@ -57,62 +57,17 @@ def import_global_config(self, path: GlobalConfigPath | None = None) -> str:
...
         self._fmu_dir.set_config_value("masterdata.smda", smda_masterdata.model_dump())
-        return f"Saved SMDA masterdata to {self._fmu_dir.path}"
+        return True, f"Saved SMDA masterdata to {self._fmu_dir.path}"
 
-    def update_model(self, model: Model) -> str:
+    def update_model(self, model: Model) -> tuple[bool, str]:
         """Save model data to the project FMU directory."""
         self._fmu_dir.set_config_value("model", model.model_dump())
-        return f"Saved model data to {self._fmu_dir.path}"
+        return True, f"Saved model data to {self._fmu_dir.path}"
```
slangeveld: Here we could just return the path (in addition to the boolean), and keep the message construction to the API-layer

src/fmu_settings_api/services/session.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        await add_fmu_project_to_session(self._session.id, fmu_dir)
+        return fmu_dir
+
+    async def close_project(self) -> tuple[bool, str]:
+        """Remove (close) a project .fmu directory from the session."""
+        project_session = cast("ProjectSession", self._session)
+        project_path = project_session.project_fmu_directory.path
+        await remove_fmu_project_from_session(self._session.id)
+        return (
```
slangeveld: Here we could just return the path (in addition to the boolean), and keep the message construction to the API-layer

slangeveld (approved):
Approving this, but think we need to agree on a pattern on what to return from the service methods before it is merged.

I am still thinking that we should try to move all the message logic to the API layer, so added a few more comments related to that. Either we keep the tuple pattern with `bool, str` or we could possibly remove the boolean altogether and only return the relevant object needed to construct the message in the API. 

I guess its mostly about making the code in the API-layer as intuitive and readable as possible, without having to read the code in the service layer to understand what is happening.

For example, from the API/route layer, if `add_access_token()` is called, I think it would be fair to expect that the token is returned (and only that), and then the Message can be constructed from that. 

`update_masterdata()` is a bit more tricky, but I guess it is fair to return the model that is updated either by reference with a path to it, or with the full `fmu_dir`object. 

Similar for the other methods in the service layer.

What do you think @mferrera?

PR #177, opened 2025-11-10
ENH: Return correct coordinate system from SMDA

### LINKED ISSUE #175: Return correct coordinate system from SMDA
The current implementation just return the first coordinate system in the list of all coordinates

https://github.com/equinor/fmu-settings-api/blob/a7658812553956b616aacb7d48f12aeb1dbd050f/src/fmu_settings_api/services/smda.py#L85-L91

It should instead return the coordinate system attached to the Field table

https://github.com/equinor/fmu-settings-api/blob/a7658812553956b616aacb7d48f12aeb1dbd050f/src/fmu_settings_api/services/smda.py#L51-L60

This data is already fetched from the Field endpoint, but is not given as the default coordinate system when the entire package is returned.

Since we can fetch multiple fields, these coordinate systems should be popped from the full list of coordinate systems and placed at the very top of the list so that they are easier to select.

GibranAlfa:
Resolves #175 

Return field-related coordinate systems at the top of the `coordinate_systems` list for easier selection in the GUI, instead of random SMDA order. 

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/smda.py
+++ b/src/fmu_settings_api/services/smda.py
@@ -86,6 +86,7 @@ class SmdaService:
         # Track seen CRS identifiers to avoid duplicates
         field_crs_list: list[CoordinateSystem] = []
         seen_field_crs_ids: set[str] = set()
+        crs_dict = {crs.identifier: crs for crs in coordinate_systems}
 
         for field in field_results:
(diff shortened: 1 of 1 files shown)


mferrera (approved):
🚀

PR #182, opened 2025-11-13
ENH: Preserve access tokens and project when renewing session

### LINKED ISSUE #179: Do not destroy session if creating a new session and one exists
https://github.com/equinor/fmu-settings-api/blob/02452a3fbc958599dc36ac1fd311ce7f9094e762/src/fmu_settings_api/v1/routes/session.py#L64-L65

A problem with this implementation is that when creating a new session, it will default to looking for the nearest .fmu directory. This is incorrect if the previous, but expired session was in a different .fmu directory.

Rather than destroying the session, it should create a new session and copy the existing state from it to the new session.

GibranAlfa:
Resolves #179 

If there's a previous session, it will create a new one and copy previous access_tokens to the new one, reattach the project to that session (if there's a project), and then destroy the old session.

## Checklist

### DIFF
--- a/src/fmu_settings_api/v1/routes/session.py
+++ b/src/fmu_settings_api/v1/routes/session.py
@@ -18,6 +18,7 @@ from fmu_settings_api.deps import (
 )
 from fmu_settings_api.models import AccessToken, Message, SessionResponse
 from fmu_settings_api.session import (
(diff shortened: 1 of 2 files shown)


mferrera (approved):
Nice

PR #192, opened 2025-11-17
ENH: Get stratigraphy units from SMDA

### LINKED ISSUE #187: Add required SMDA interface methods for pulling stratigraphic data
SMDA has some API routes for stratigraphic data. We need to support these to pull the right data from SMDA.

GibranAlfa:
Resolves #187 

- Add POST /strat-units to get the stratigraphy units from SMDA. The new `StratigraphicUnit` model is stolen from Dynageo: https://github.com/equinor/dynageo/blob/c26d2091d0b4a9d864bec6492c69e00cb5442159/dynageo/data_validation/smda_validation.py#L446-L466
- This `StratigraphicUnit` model should probably in fmu-datamodels?
- There are other columns we can get from SMDA, but I omitted them to follow what Dynageo did. (ref to https://opus.smda.equinor.com/smda_viewer/strat-column )
- This is not validated like in Dynageo because I don't know what are they validating for (hard to see what is their intention in the code, maybe just me).
- docstring in `StratigraphicUnit` model not updated yet, just want to see some comments first.
- `uv.lock` is the updated lock from running `uv sync --all-extras --upgrade`. this is okay right?

## Checklist

### DIFF
--- a/src/fmu_settings_api/models/smda.py
+++ b/src/fmu_settings_api/models/smda.py
@@ -1,6 +1,6 @@
 """Models (schemas) for the SMDA routes."""
 
-from typing import Annotated, Literal
+from typing import Literal
 from uuid import UUID
 
 from fmu.datamodels.fmu_results.fields import (
@@ -14,16 +14,19 @@ from pydantic import Field
 
 from fmu_settings_api.models.common import BaseResponseModel
 
-# Pydantic custom types
-StratLevelInt = Annotated[int, Field(ge=1, le=6)]
-NonNanNonNegativeFloat = Annotated[float, Field(ge=0, allow_inf_nan=False)]
-
 
 class SmdaField(BaseResponseModel):
     """An identifier for a field to be searched for."""
 
     identifier: str = Field(examples=["TROLL"])
     """A field identifier (name)."""
+
+
+class SmdaStratColumn(BaseResponseModel):
+    """An identifier for a stratigraphic column."""
+
+    strat_column_identifier: str = Field(examples=["LITHO_TROLL"])
+    """A stratigraphic column identifier."""
 
 
 class SmdaFieldUUID(BaseResponseModel):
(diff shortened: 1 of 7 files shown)


ErichSuter (approved):
LGTM 👍

src/fmu_settings_api/models/smda.py
```diff
@@ -67,3 +72,39 @@ class SmdaMasterdataResult(BaseResponseModel):
...
+    strat_unit_parent: str | None
+    strat_column_type: Annotated[
+        str,
+        (
+            Literal["lithostratigraphy"]
+            | Literal["sequence stratigraphy"]
+            | Literal["chronostratigraphy"]
+            | Literal["biostratigraphy"]
+        ),
+    ]
```
mferrera: I think you can just do `Literal["lithostratigraphy", "sequence ...", ...]`. A bit less verbose.

src/fmu_settings_api/models/smda.py
```diff
@@ -13,6 +14,10 @@
 
 from fmu_settings_api.models.common import BaseResponseModel
 
+# Pydantic custom types
```
mferrera: I don't think this comment is necessary

src/fmu_settings_api/v1/routes/smda/main.py
```diff
@@ -210,3 +211,74 @@ async def post_masterdata(
             detail=str(e),
             headers={HttpHeader.UPSTREAM_SOURCE_KEY: HttpHeader.UPSTREAM_SOURCE_SMDA},
         ) from e
+
+
+@router.post(
+    "/strat-units",
```
mferrera: This would be our first route on the API using a hyphen instead of an underscore. Suggest keeping with the existing theme and using `strat_units` (even if SMDA uses the hyphen). Depending on what @tnatt / @joargr prefer, maybe it should even be the verbose `stratigraphic_units`.

src/fmu_settings_api/interfaces/smda_api.py
```diff
@@ -115,6 +116,19 @@ async def strat_column_areas(
...
+    async def strat_units(
+        self, field_identifiers: Sequence[str], columns: Sequence[str] | None = None
+    ) -> httpx.Response:
+        """Searches for the stratigraphic units related to a field identifier."""
+        _projection = "identifier,uuid" if columns is None else ",".join(columns)
+        return await self.post(
+            SmdaRoutes.STRAT_UNITS_SEARCH,
+            json={
+                "_projection": _projection,
+                "strat_column_identifier": field_identifiers,
```
tnatt: Might be my lack of knowledge about the smda API 🙂 but does this actually return results if we use field identifiers as the `strat_column_identifier` ?
Should we not be using the stratigraphic column identifier here instead and pass the selected stratigraphic column? Looking at dynageo they seem to provide the field identifier separately in addition to the `strat_column_identifier`.

In the frontend we should only display units from the selected stratigraphic column, and I don't see a way of filtering the results based on the stratigraphic column they belong to from data in the returned response.

GibranAlfa: oops yes I overlooked this. I tried with Troll but they have the same field identifiers and `strat_column_identifier`. will fix this, thanks

PR #197, opened 2025-11-21
ENH: Add endpoints to find and set RMS projects

### LINKED ISSUE #186: Add end point for accessing list of RMS models in project
This will return a list of RMS models to the GUI. This will allow selecting an RMS project to configure in the FMU project, or, to extract information from (i.e. stratigraphy.)

This is basically a wrapper around https://github.com/equinor/fmu-settings/issues/112

In the typical case there is only one. But this isn't a guarantee.

GibranAlfa:
Resolves #186 

- Added GET `/rms_projects` to get a list of paths to RMS projects found in the current project directory.
- Added PATCH `/rms_project_path` to change the RMS project path in config file

I think the endpoint naming is fine but if you have any preferences just tell me.

Long lines of code are mostly tests and `uv.lock`

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/__init__.py
+++ b/src/fmu_settings_api/services/__init__.py
@@ -1,7 +1,6 @@
 """Services the end points need to use."""
 
 from fmu_settings_api.services.project import ProjectService
-from fmu_settings_api.services.session import SessionService
 from fmu_settings_api.services.smda import SmdaService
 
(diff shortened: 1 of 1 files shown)


GibranAlfa:
I refactored other methods to only return bool as well, accessing the path can be done via path property

PR #213, opened 2025-12-02
ENH: Delete stale lock when checking lock status

### LINKED ISSUE #205: Checking lock status should delete an expired lock file
There is an issue in the GUI now after the lock and the session length is decoupled. 
The GUI can end up in a situation where the project has turned read-only but the lock information is still available - hence the user will not be prompted that the project can be opened for editing.

The GUI checks the lock status every nth minute, and I suspect that the only change that needs to happen is that the lock file is deleted if the lock has expired. This should trigger the expected GUI behavior.

GibranAlfa:
Resolves #205 

get_lock_status will delete stale lock now

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/session.py
+++ b/src/fmu_settings_api/services/session.py
@@ -109,6 +109,16 @@ class SessionService:
                 lock_file_exists = True
                 try:
                     lock_info = fmu_dir._lock.load(force=True, store_cache=False)
+                    if fmu_dir._lock._is_stale(lock_info):
+                        try:
+                            fmu_dir._lock.path.unlink()
+                            lock_file_exists = False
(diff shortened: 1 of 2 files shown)


src/fmu_settings_api/services/session.py
```diff
@@ -109,6 +109,16 @@ def get_lock_status(self) -> LockStatus:
                 lock_file_exists = True
                 try:
                     lock_info = fmu_dir._lock.load(force=True, store_cache=False)
+                    if fmu_dir._lock._is_stale(lock_info):
```
GibranAlfa: how likely it is that someone acquire the lock after reload the lock_info here? i think this is okay

PR #214, opened 2025-12-02
MAINT: Add exception handler for generic 500 errors

### LINKED ISSUE #204: Add exception handler for generic 500 errors
FastAPI allows specific exception handlers to uniformly handle certain exceptions.

```python
@app.exception_handler(Exception)
async def unhandled_exception_handler(request: Request, exc: Exception) -> JSONResponse:
    """Handle unexpected errors."""
    logger.exception(
        "Unexpected error occurred",
        extra={
            "path": request.url.path,
            "error": str(exc),
            "error_type": type(exc).__name__,
        }
    )
    
    return JSONResponse(
        status_code=500,
        content={
            "detail": "An unexpected error occurred.",
        },
    )
```

This will ensure unhandled/unexpected exceptions have their details _logged_ as an _event_ with structlog, rather than just sending the exception string back to the user. And then the catch-all `except Exception as e` should not need to be present on every route.

GibranAlfa:
Resolves #204

## Checklist

### DIFF
--- a/src/fmu_settings_api/middleware/logging.py
+++ b/src/fmu_settings_api/middleware/logging.py
@@ -51,7 +51,17 @@ class LoggingMiddleware(BaseHTTPMiddleware):
                 duration_ms=round(duration * 1000, 2),
             )
             return response
-        except HTTPException:
+        except HTTPException as e:
+            duration = time.time() - start_time
+            logger.warning(
+                "request_failed",
+                method=request.method,
+                path=request.url.path,
+                status_code=e.status_code,
+                error=e.detail,
+                error_type=type(e).__name__,
+                duration_ms=round(duration * 1000, 2),
+            )
             raise
         except Exception as e:
             duration = time.time() - start_time


src/fmu_settings_api/middleware/logging.py
```diff
@@ -59,4 +63,7 @@ async def dispatch(
                 error_type=type(e).__name__,
                 duration_ms=round(duration * 1000, 2),
             )
-            raise
+            return JSONResponse(
+                status_code=500,
+                content={"detail": "An unexpected error occurred."},
+            )
```
mferrera: Does this need to be added? It prevents logging request failures for HTTPExceptions, doesn't it? Can you say a bit about the interaction between this and the exception handler?

GibranAlfa: yes, because this middleware catch exception before the app level handler catch it, and if we just raise the error here, it will crash and raise the internal error (which is not what we want, this should propagate to app level error but this maybe the BaseHTTPMiddleware limitation).

turns out the app level handler is never executed, so we could actually remove it and it will behave properly.

PR #219, opened 2025-12-04
ENH: Add PATCH endpoints to set the new fields in RmsProject model

### LINKED ISSUE #217: Consider updating the PATCH /project/rms endpoint with the new RmsProject model
investigate if we should update it or make new endpoints for the optional entry in the RmsProject model in ProjectConfig

GibranAlfa:
Resolves #217 

- Added new patch endpoints to set the coordinate systems, wells, horizons, and zones in the config file
- Remove the previous API response model and use the model from fmu.settings instead

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dynamic = ["version"]
 dependencies = [
     "fastapi",
     "fmu-datamodels",
-    "fmu-settings>=0.11.0",
+    "fmu-settings>=0.13.0",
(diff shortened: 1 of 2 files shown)


GibranAlfa:
i'm thinking making a granular routes since we only do strat mappings for now, and these endpoints can be called on the save button in strat mappings page for example. but we can do it with just patch /rms as well

we can discuss this further on planning meeting

PR #220, opened 2025-12-08
ENH: Add `MatchService` to match two different entities

### LINKED ISSUE #191: Match RMS project coordinate system to SMDA coordinate system
Now that we have the ability to open an RMS project and retrieve data from it, we should be able to match the coordinate system used in the project to an SMDA coordinate system.

This should allow us extra confidence in auto-populating the coordinate system.

### LINKED ISSUE #218: Add assignment/matching algorithm for stratigraphy matching
We want to present the GUI with an already mostly-matched stratigraphy between SMDA and RMS.

scipy has the Hungarian algorithm as https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linear_sum_assignment.html.

We could consider copying that (or a similar optimization function is contains) into here, or using it directly.

A cost function should be evaluated that weighs:

- Age (for vertical hierarchy)
- Similarity (text matching)

GibranAlfa:
Resolves #218 
Resolves #191 

(254 line counts from uv.lock)

- Added a new `/match` route with `GET /stratigraphy` and `GET /coordinate_system` which return the results of matching RMS and SMDA stratigraphy and coordinate system respectively. Any inputs for naming or arranging the routes are welcomed.
- Matching coordinate system is weird because we have one value from RMS and SMDA, so I guess RMS coordinate system just need to be mapped to SMDA coordinate system?
- The models for matching is specified to match RMS to SMDA for now, but we can generalize later if we need to match from other source/target, because it's not clear now what's the schema for other source/target
- Similarity Score is counted using Levenshtein distance, a string metric that measures the difference between two sequences by calculating the minimum number of single-character edits (insertions, deletions, or substitutions) needed to change one word into the other. The lower the edits needed, the higher the similarity score. Strict name score calculate similarity with exact word order, while the other one allow for different word ordering (examples are in docstring).
- We only calculate name similarity for now. Using horizon name is not helpful as proven from experiment on JS and Troll.
- We have confidence level which is determined by the score. The idea is the GUI can highlight the results that have low confidence level and ask the users to double check. But as we can also have wrong results on high confidence level, maybe this metric is not that useful and the GUI should ask the users to review everything anyway.
- Tests are not added yet as the implementation can change after review.

## Checklist


src/fmu_settings_api/services/match.py
```diff
@@ -0,0 +1,264 @@
...
+            strat_column_identifier
+        )
+
+        return self.match_stratigraphy(rms_zones, smda_units_result.stratigraphic_units)
+
+    def match_coordinate_system_from_config(
+        self,
+        project_session: "ProjectSession",
+    ) -> RmsCoordinateSystemMatch:
+        """Match RMS coordinate system to SMDA coordinate system from project config.
```
slangeveld: Same with this docstring as above

src/fmu_settings_api/services/match.py
```diff
@@ -0,0 +1,264 @@
...
+    from fmu_settings_api.session import ProjectSession
+
+HIGH_CONFIDENCE_THRESHOLD = 80
+MEDIUM_CONFIDENCE_THRESHOLD = 50
+
+
+class MatchService:
+    """Service for matching two different entities."""
+
+    async def match_stratigraphy_from_config(
```
slangeveld: As this matching service seems to be quite general (not SmdaMatchingService), maybe the methods that are actually covering SMDA matching, should have `smda` in their names?

slangeveld (commented):
Nice work! Easy to read and well structured. A few comments on namings and response objects.

Also, remember to add tests for the new endpoints and the MatchService 🧪

slangeveld:
Remember to add tests for the new endpoints and the MatchService 🧪

src/fmu_settings_api/services/match.py
```diff
@@ -0,0 +1,264 @@
...
+                "No SMDA coordinate system found in project masterdata. "
+                "Please configure masterdata.smda.coordinate_system "
+                "in the project config."
+            )
+
+        smda_crs = CoordinateSystem.model_validate(smda_crs_config)
+
+        return self.match_coordinate_system(rms_crs, smda_crs)
+
+    def match_stratigraphy(
```
slangeveld: Maybe these method names should also be adjusted if it only RMS/SMDA matching that is done

src/fmu_settings_api/services/match.py
```diff
@@ -0,0 +1,264 @@
...
+                        rms_zone=zone,
+                        smda_unit=best_unit,
+                        score=best_score,
+                        confidence=confidence,
+                    )
+                )
+
+        return matches
+
+    def match_coordinate_system(
```
slangeveld: Same with this method name

PR #240, opened 2026-01-12
ENH: Support opening rms projects with version

### LINKED ISSUE #228: Give proper response for rmsapi.LicenseError
When there is an issue for checking out an rmsapi license an internal server error is raised.
We should have a proper error response for these, so that it can be displayed in the gui why RMS was unable to open.

Image

### LINKED ISSUE #239: Support opening RMS projects with parameterized version
In the route that opens an RMS project, we should allow the GUI to send an RMS version to open the RMS project with.

The use case is that RMS API features in newer versions may provide additional data or context we need. For example, RMS 15+ stratigraphy data contains additional stratigraphic column data useful for create a thorough mapping.

At least two different failure states with this: the version does not exist/is not valid, and the project cannot be opened with that version. It is unusual for the latter to the be case, but it is possible, for example, if trying to open an RMS project 12.1.4 directly to 15, it may attempt and upgrade migration that can fail. (In practice, this is unlikely now, as we can only support reading RMS 14+. But in the future it could be a problem).

There are some other conditions worth considering: what if the requested version is _lower_ than the current version? (I don't have an answer for this)

slangeveld:
Resolves #239 
Resolved #228 

Support opening rms project with specified version.

When opening an RMS project with the POST endpoing `api/v1/rms/`, there is now an option to specify the RMS version the project should be opened with. 

Opening an RMS project can fail for different reasons. The following issues are catched and returned to the user with an `HTTPException`:
* Your project does not support this version of RMS.
* Unable to check out required license.
* Failed setting up RMS API proxy. The RMS version is invalid.

Also refactored and generalized the exception handling a bit and moved all responses to the new `FailedOpeningRmsProjectResponses` object.

## Checklist

### DIFF
--- a/src/fmu_settings_api/models/rms.py
+++ b/src/fmu_settings_api/models/rms.py
@@ -24,5 +24,5 @@ class RmsProjectPathsResult(BaseResponseModel):
 class RmsVersion(BaseResponseModel):
     """RMS version."""
 
-    version: str = Field(examples=["14.2.2", "15..1.0"])
+    version: str = Field(examples=["14.2.2", "15.0.1.0"])
     """A version of RMS."""
--- a/src/fmu_settings_api/v1/routes/rms.py
+++ b/src/fmu_settings_api/v1/routes/rms.py
@@ -10,9 +10,8 @@ from fmu.settings.models.project_config import (
     RmsStratigraphicZone,
     RmsWell,
 )
-from runrms.exceptions import RmsProjectNotFoundError
 from runrms.api.proxy import RemoteException
-from runrms.exceptions import RmsVersionError
+from runrms.exceptions import RmsProjectNotFoundError, RmsVersionError
 
 from fmu_settings_api.deps import SessionServiceDep
 from fmu_settings_api.deps.rms import (
@@ -54,14 +53,39 @@ RmsResponses: Final[Responses] = {
 
 FailedOpeningRmsProjectResponses: Final[Responses] = {
     **inline_add_response(
-        422,
-        dedent("Failed opening RMS project."),
+        400,
+        dedent("RMS project path is not configured in the project config file."),
         [
-            {"detail": "Your project does not support this version of RMS."},
-            {"detail": "Unable to check out required license."},
-            {"detail": "Failed setting up RMS API proxy. The RMS version is invalid. "},
+            {"detail": "RMS project path is not set in the project config file."},
         ],
-    )
+    ),
+    **inline_add_response(
+        404,
+        dedent("RMS project does not exist at the configured path."),
+        [
+            {
+                "detail": (
+                    "RMS project does not exist at the "
+                    "configured path: {rms_project_path}."
+                )
+            },
+        ],
+    ),
+    **inline_add_response(
+        422,
+        dedent("Failed to open RMS project {rms_project_path}."),
+        [
+            {"detail": "Could not open project using RMS version {version}."},
+            {"detail": "Unable to check out required license."},
+            {
+                "detail": (
+                    "Failed setting up RMS API proxy: The requested RMS version "
+                    "{version} is not supported. Try specifying another RMS version "
+                    "or upgrade the RMS project."
+                )
+            },
+        ],
+    ),
 }
 
 router = APIRouter(prefix="/rms", tags=["rms"])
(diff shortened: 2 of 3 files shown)


src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -48,45 +51,103 @@
...
+            )
+        elif "Unable to check out required license." in str(e):
+            error_msg = (
+                "Failed opening RMS project. Unable to check out"
+                f"required license: {str(e)}."
+            )
+        else:
+            error_msg = f"Failed opening RMS project: {str(e)}"
+        raise HTTPException(status_code=422, detail=error_msg) from e
+    except RmsVersionError as e:
```
slangeveld: With an invalid RMS version it fails before the RMS API Proxy is created so this does not raise an RemoteException. Here we need to catch the RmsVersionError instead.

src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -48,45 +51,103 @@
...
+                f"of RMS: {str(e)}."
+                ""
+            )
+        elif "Unable to check out required license." in str(e):
+            error_msg = (
+                "Failed opening RMS project. Unable to check out"
+                f"required license: {str(e)}."
+            )
+        else:
+            error_msg = f"Failed opening RMS project: {str(e)}"
```
slangeveld: This error message is supposed to catch the other RemoteException cases that are not identified yet.

slangeveld:
@tnatt (and others) Suggested improvements for the error messages sent to the GUI are very much welcomed

src/fmu_settings_api/models/rms.py
```diff
@@ -19,3 +19,10 @@ class RmsProjectPathsResult(BaseResponseModel):
 
     results: list[RmsProjectPath]
     """List of absolute paths to RMS projects within the FMU project."""
+
+
+class RmsVersion(BaseResponseModel):
+    """RMS version."""
+
+    version: str = Field(examples=["14.2.2", "15..1.0"])
```
tnatt: Typo?
```suggestion
    version: str = Field(examples=["14.2.2", "15.0.1.0"])
```

src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -48,45 +51,103 @@
...
+        raise HTTPException(status_code=422, detail=error_msg) from e
```
tnatt: I think it could be useful for the users to see the rms project path for all of these error messages.
Maybe not the version, since it will just cause confusion where we open the project in a different version than their project.

src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -48,45 +51,103 @@
...
+        root_proxy, project = rms_service.open_rms_project(rms_project_path, version)
         await session_service.add_rms_session(root_proxy, project)
         return Message(
-            message=f"RMS project opened successfully with RMS version {rms_version}"
+            message=f"RMS project opened successfully with RMS version {version}."
         )
     except SessionNotFoundError as e:
         raise HTTPException(status_code=401, detail=str(e)) from e
+    except RemoteException as e:
+        if "File version" in str(e) and "is not supported" in str(e):
```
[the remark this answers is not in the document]
slangeveld: Did almost as suggested, but just to have it similar to the other error messages I ended up doing:
`Failed to open RMS project {rms_project_path}: Could not open project using RMS version {version}`

src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -49,57 +51,115 @@
...
-            message=f"RMS project opened successfully with RMS version {rms_version}"
+            message=f"RMS project opened successfully with RMS version {version}."
         )
     except SessionNotFoundError as e:
         raise HTTPException(status_code=401, detail=str(e)) from e
     except RmsProjectNotFoundError as e:
         raise HTTPException(
-            status_code=404,
-            detail=f"RMS project not found at path: {rms_project_path}",
+            status_code=422,
```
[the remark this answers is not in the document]
GibranAlfa: just noticed the test_open_rms_project_not_found is not included here (https://github.com/equinor/fmu-settings-api/commit/f45acce515724739023d2407b079c646d31850bd), this should cause merge conflict though 🤔 , strange

tnatt (approved):
One small comment, other than that this LGTM 👍

PR #253, opened 2026-01-13
ENH: Add cache endpoints

### LINKED ISSUE #185: Add endpoint for accessing the cache
We need a way to access and restoring from the cache from the front end. 

The GUI should be able to:
 - restore a project that has become corrupt e.g. due to manually editing of the files, to a previously working state. 
 - undo recent changes to the project (through undo buttons)

Not entirely sure how the endpoint should behave, so it requires a little group discussion.

GibranAlfa:
Resolves #185  

In project route:
- Added `GET /cache` to list available caches/revisions of a specific resource. Right now we only have `config.json` as cached resource in project .fmu
- Added `GET /cache/{revision_id}` to view the content of the cache. Revision_id is the filename of the cache, returned by the `GET /cache`
- Added `POST /cache/restore/{revision_id}` to restore a resource from a cache revision by overwriting current resource with the cache. The current resource state is cached before overwriting (when present). Example for when a resource is not present is no current config file (deleted or missing)
**Example flow:**

        1. Current resource is in state A
        2. Call restore with a revision from state B
        3. Your current state A is cached (when present)
        4. Resource is now in state B
        5. To undo: restore from the newly created cache entry (your state A backup)

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/resource.py
+++ b/src/fmu_settings_api/services/resource.py
@@ -51,7 +51,13 @@ class ResourceService:
 
         try:
             self._fmu_dir.restore_from_cache(resource_path, revision_id)
-        except (FileNotFoundError, ValueError):
+        except (FileNotFoundError, ValueError) as e:
+            logger.error(
+                "cache_restore_failed",
+                resource=resource.value,
+                revision_id=revision_id,
+                error=str(e),
+            )
             raise
 
         logger.info(


src/fmu_settings_api/services/resource.py
```diff
@@ -0,0 +1,114 @@
...
+        """Returns the path to the .fmu directory."""
+        return self._fmu_dir.path
+
+    def list_cache_revisions(self, resource: CacheResource) -> CacheList:
+        """List all cache revisions for a specific resource from oldest to newest."""
+        resource_path = Path(resource.value)
+        revision_paths = self._fmu_dir.cache.list_revisions(resource_path)
+        return CacheList(revisions=[path.name for path in revision_paths])
+
+    def get_cache_content(
```
slangeveld: Could the logic in this method me moved to `fmu-settings`? There seems to be quite some logic here that is more related to the files and structure of .fmu which the API layer should not know about, so that should probably be moved to there.

slangeveld (dismissed):
Most of this looks good, but it feels like some of the logic in the ResourceService regarding the cache should be moved to the CacheManager in `fmu-settings`. Logic related to the files and structure of .fmu is not something the API layer should know (or care) about. By adding this to the API layer we add more dependencies between the layers that is needed.

The API layer should use the resource managers in `fmu-settings` as interface towards the files and folders in .fmu. In that way we can easily change the structure and logic of `.fmu`, without having to make any changes in the API layer.

GibranAlfa:
> Most of this looks good, but it feels like some of the logic in the ResourceService regarding the cache should be moved to the CacheManager in `fmu-settings`. Logic related to the files and structure of .fmu is not something the API layer should know (or care) about. By adding this to the API layer we add more dependencies between the layers that is needed.
> 
> The API layer should use the resource managers in `fmu-settings` as interface towards the files and folders in .fmu. In that way we can easily change the structure and logic of `.fmu`, without having to make any changes in the API layer.

agreed, will make an issue in fmu-settings to move some functionality to the CacheManager. but is there any comments regarding the logic or the flow here?

src/fmu_settings_api/models/resource.py
```diff
@@ -0,0 +1,19 @@
...
+    """List of cache revision filenames."""
+
+    revisions: list[str]
+    """Cache revision filenames, sorted oldest to newest."""
+
+
+class CacheContent(BaseResponseModel):
+    """Content of a cache revision."""
+
+    data: dict[str, Any]
```
mferrera: It's technically possible for these files to be any valid JSON, which has a much more complicated type signature. Borrowing from elsewhere:

```python
JsonValue: TypeAlias = (
    int | float | str | bool | None | list["JsonValue"] | dict[str, "JsonValue"]
)
```

I don't propose we change this now, but it may hit us at some point. It could be a restriction we place on ourselves that all Pydantic resources as files are structured as mappings (dictionaries) which is likely to be the case anyway. (Just a note for thought)

GibranAlfa: was thinking about this as well but I think all of our resources will be dictionaries anyway

PR #263, opened 2026-01-16
ENH: Fetch stratigraphic column and horizon type from RMS

### LINKED ISSUE #223: Add a helper method to fetch stratigraphic columns from an RMS project
Depends on https://github.com/equinor/fmu-settings/issues/143 

this helper method should be used in `get_zones()` in `RmsService`, so that it will return a list of `RmsStratigraphicZone` with the newly added `stratigraphic_column` field

### LINKED ISSUE #259: Extract horizon type from RMS
It is possible to extract the type of an RMS horizon. The `RmsService.get_horizons()` should set this `type` once this update of the model has been completed: https://github.com/equinor/fmu-settings/issues/151

It works to extract the type in RMS v14 and does not need to be optional.

### LINKED ISSUE #262: Handle detached zones in RMS
It is possible in RMS to store a Stratigraphic Framework with detached zones (annoyingly). 

Stratigraphic Framework to the right 👇 was made by deleting a the **UJUNC3** horizon that constrained the base of the **MiddleLowerDraupne** zone and the top of the **UpperDraupne.**

This was possible to save in RMS. When trying to fetch the data through the GUI afterwards I got a 500 respons. Seems like the zones could be stored to RMS but one of their `horizon_above`/`horizon_below` attributes are `None` leading to `AttributeError: 'NoneType' object has no attribute 'name'`.

We should probably just ignore these detached zones as it is an invalid configuration that RMS should not allow, this can be done by skipping zones where one (or both) of the horizons are `None`

| Before deleting UJUNC3| After deleting UJUNC3|
|--|--|
| Image|Image|

GibranAlfa:
1. Resolves #262 
2. Resolves #259 
3. Resolves #223 

Added RmsVersion dependency so when we open an RMS project with a specified version, the get_zones() will automatically use that RMS version.

Output for Drogon model:
```
[
  {
    "name": "Above",
    "top_horizon_name": "MSL",
    "base_horizon_name": "TopVolantis",
    "stratigraphic_column_name": [
      "Column",
      "Column_1"
    ]
  },
  {
    "name": "Valysar",
    "top_horizon_name": "TopVolantis",
    "base_horizon_name": "TopTherys",
    "stratigraphic_column_name": [
      "Column"
    ]
  },
  {
    "name": "Therys",
    "top_horizon_name": "TopTherys",
    "base_horizon_name": "TopVolon",
    "stratigraphic_column_name": [
      "Column"
    ]
  },
  {
    "name": "Volon",
    "top_horizon_name": "TopVolon",
    "base_horizon_name": "BaseVolantis",
    "stratigraphic_column_name": [
      "Column"
    ]
  },
  {
    "name": "Below",
    "top_horizon_name": "BaseVolantis",
    "base_horizon_name": "BaseVelmodel",
    "stratigraphic_column_name": [
      "Column",
      "Column_1"
    ]
  },
  {
    "name": "TopVolantis_BaseVolantis",
    "top_horizon_name": "TopVolantis",
    "base_horizon_name": "BaseVolantis",
    "stratigraphic_column_name": [
      "Column_1"
    ]
  }
]
```

```
[
  {
    "name": "MSL",
    "type": "interpreted"
  },
  {
    "name": "Seabase",
    "type": "interpreted"
  },
  {
    "name": "TopVolantis",
    "type": "interpreted"
  },
  {
    "name": "TopTherys",
    "type": "interpreted"
  },
  {
    "name": "TopVolon",
    "type": "interpreted"
  },
  {
    "name": "BaseVolantis",
    "type": "interpreted"
  },
  {
    "name": "BaseVelmodel",
    "type": "interpreted"
  }
]
```

## Checklist


src/fmu_settings_api/services/rms.py
```diff
@@ -44,22 +44,56 @@ def open_rms_project(
...
+            ]
+        # RMS 14 and earlier don't support stratigraphic columns
         return [
             RmsStratigraphicZone(
                 name=zone.name.get(),
                 top_horizon_name=zone.horizon_above.name.get(),
                 base_horizon_name=zone.horizon_below.name.get(),
             )
             for zone in rms_project.zones
+            if zone.horizon_above is not None and zone.horizon_below is not None
```
GibranAlfa: can't use `for zone in rms_project.zones if (zone.horizon_above and zone.horizon_below)` because zone object doesn't support `__bool__`

src/fmu_settings_api/services/rms.py
```diff
@@ -71,7 +105,17 @@ def get_horizons(self, rms_project: RmsApiProxy) -> list[RmsHorizon]:
         Returns:
             list[RmsHorizon]: List of horizons in the project
         """
-        return [RmsHorizon(name=horizon.name.get()) for horizon in rms_project.horizons]
+        horizons = []
+        for horizon in rms_project.horizons:
+            horizon_type = horizon.type.get()
+            type_str = str(horizon_type).split(".")[-1]
```
GibranAlfa: `horizon.type.get()` will return `_rmsapi.HorizonType.interpreted` for example, so just need to take the last string

src/fmu_settings_api/services/rms.py
tnatt: I would simply remove a horizon, press apply and save the project as a test with a new name using Save as 👍

mferrera: `__bool__` is implemented in the proxy, but I think it has a bug:

```python
def __bool__(self) -> int: # oops
```

tnatt:
> so calling `if zone.horizon_above.get() and zone.horizon_below.get():` will return `AttributeError: 'Horizon' object has no attribute '__bool__'`, suggesting RMS horizon object has no `__bool__`.
> 
> calling `if (zone.horizon_above.get() is not None and zone.horizon_below.get() is not None):` works and will only return valid zones.

Nice! That is much better than the suppressing the exception 🙂

src/fmu_settings_api/services/rms.py
```diff
@@ -44,23 +46,43 @@ def open_rms_project(
         rms_proxy = get_rmsapi(version=rms_version)
         return rms_proxy, rms_proxy.Project.open(str(rms_project_path), readonly=True)
 
-    def get_zones(self, rms_project: RmsApiProxy) -> list[RmsStratigraphicZone]:
+    def get_zones(
+        self, rms_project: RmsApiProxy, rms_version: str
```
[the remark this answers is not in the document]
tnatt: We might be talking past each other here..  but isn't this what we want? i.e. we want to use the version that the project have been opened in? 
Say my project is actually in v14, but the GUI will try to open it with v15, and if it succeeds it should use this version for get_zones.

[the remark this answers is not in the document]
GibranAlfa: yes, it shows 1.11 for rms14.2.2 and 1.13 for rms15.0.1.0, so which one should we use? is it possible that the API version change in a same rms version?

src/fmu_settings_api/services/rms.py
```diff
@@ -53,14 +55,33 @@ def get_zones(self, rms_project: RmsApiProxy) -> list[RmsStratigraphicZone]:
...
-                name=zone.name.get(),
-                top_horizon_name=zone.horizon_above.name.get(),
-                base_horizon_name=zone.horizon_below.name.get(),
-            )
-            for zone in rms_project.zones
-        ]
+        zone_columns: dict[str, list[str]] = {}
+        rms_api_version = str(getattr(rms_project, "__version__", "0"))
+        major_minor = tuple(int(part) for part in rms_api_version.split(".")[:2])
+        if major_minor >= MIN_RMS_API_VERSION_FOR_STRAT_COLUMNS:
```
tnatt: suggest using the Version class here from `packaging.version.` And do we need the `getattr` or could we access it directly?

```python
from packaging.version import Version

MIN_RMS_API_VERSION_FOR_STRAT_COLUMNS = Version("1.12")

rms_api_version = rms_project.__version__  
if Version(rms_api_version) >= MIN_RMS_API_VERSION_FOR_STRAT_COLUMNS:
```

PR #294, opened 2026-02-13
ENH: Add changelog endpoint

### LINKED ISSUE #184: Add endpoint for accessing the changelog
We should implement in the GUI a way to display the changelog with filter possibilities, but currently we don't have any way of accessing the changelog from the front end. This could be provided through a dedicated endpoint in the API. 

Initially we don't need a way to filter the changelog, that can be performed in the browser, and could be added later to the endpoint itself to only get a subset of the changes in return.

nourinmohd:
Resolves #184 

Added the endpoint for changelog. 

## Checklist

### DIFF
--- a/src/fmu_settings_api/v1/routes/project.py
+++ b/src/fmu_settings_api/v1/routes/project.py
@@ -275,13 +275,8 @@ CacheResponses: Final[Responses] = {
 ChangelogResponses: Final[Responses] = {
     **inline_add_response(
         404,
-        "Changelog resource file not found",
+        "Changelog file not found",
         [{"detail": "No changelog file found at {path}"}],
-    ),
-    **inline_add_response(
-        403,
-        "Permission denied accessing changelog file",
-        [{"detail": "Permission denied while trying to access changelog at {path}"}],
     ),
     **inline_add_response(
         422,
@@ -1160,31 +1155,6 @@ async def put_mappings(
         raise HTTPException(status_code=400, detail=str(e)) from e
 
 
-def _create_opened_project_response(fmu_dir: ProjectFMUDirectory) -> FMUProject:
-    """Creates an FMUProject response model for an opened project.
-
-    Includes path, configuration, and read-only status determined by lock acquisition.
-    Raises HTTP exceptions for corrupt or inaccessible projects.
-    """
-    try:
-        service = ProjectService(fmu_dir)
-        return service.get_project_data()
-    except FileNotFoundError as e:
-        raise HTTPException(
-            status_code=404,
-            detail=f"Missing configuration file in project at {fmu_dir.path}: {e}",
-        ) from e
-    except ValueError as e:
-        raise HTTPException(
-            status_code=422,
-            detail=f"Corrupt project found at {fmu_dir.path}: {e}",
-        ) from e
-    except PermissionError as e:
-        raise HTTPException(
-            status_code=403, detail="Permission denied accessing .fmu"
-        ) from e
-
-
 @router.get(
     "/changelog",
     response_model=Log[ChangeInfo],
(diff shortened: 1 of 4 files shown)


src/fmu_settings_api/v1/routes/project.py
```diff
@@ -269,6 +272,34 @@
...
+ChangelogResponses: Final[Responses] = {
+    **inline_add_response(
+        404,
+        "Changelog resource file not found",
+        [{"detail": "No changelog file found at {path}"}],
+    ),
+    **inline_add_response(
+        403,
+        "Permission denied accessing changelog file",
+        [{"detail": "Permission denied while trying to access changelog at {path}"}],
```
GibranAlfa: this is already covered by the projectresponse so no need to add it here

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -1220,11 +1213,3 @@ async def get_changelog(
             status_code=403,
             detail="Permission denied while trying to read the changelog.",
         ) from e
-    except ValueError as e:
```
GibranAlfa: still need to catch valueerror

tests/test_v1/test_project.py
```diff
@@ -296,6 +296,52 @@ async def test_get_project_already_in_session(
     assert session.project_fmu_directory.config.load() == fmu_project.config
 
 
+async def test_get_changelog_success(
+    client_with_session: TestClient,
+) -> None:
+    """Test 200 is returned when changelog exists and is readable."""
+    response = client_with_session.get(ROUTE)
```
GibranAlfa: this is not calling to get changelog, it should be `response = client_with_project_session.get(f"{ROUTE}/changelog")`

we also need to create a changelog entry first and assert the result. see `GET project/cache` test and follow the pattern.

create a project session
```python
    session_id = client_with_project_session.cookies.get(
        settings.SESSION_COOKIE_KEY, None
    )
    assert session_id is not None
    session = await session_manager.get_session(session_id)
    assert isinstance(session, ProjectSession)
```
create a changelog entry, for example
```python
    fmu_dir = session.project_fmu_directory
    fmu_dir.changelog.add_log_entry(
        ChangeInfo(
            change_type=ChangeType.update,
            user="tester",
            path=fmu_dir.path,
            change="Updated field 'cache_max_revisions'",
            hostname="localhost",
            file="config.json",
            key="cache_max_revisions",
        )
    )
```
call the endpoint and assert the result
```python
    response = client_with_project_session.get(f"{ROUTE}/changelog")
    assert response.status_code == status.HTTP_200_OK

    response_data = response.json()
    assert len(response_data) == 1
    assert response_data[0]["change_type"] == "update"
    assert response_data[0]["file"] == "config.json"
    assert response_data[0]["key"] == "cache_max_revisions"
```

tests/test_v1/test_project.py
```diff
@@ -296,6 +296,52 @@ async def test_get_project_already_in_session(
...
+) -> None:
+    """Test 404 is returned when changelog file does not exist."""
+    # Remove .fmu directory to simulate missing changelog
+    fmu_dir = session_tmp_path / ".fmu"
+    if fmu_dir.exists():
+        for item in fmu_dir.iterdir():
+            if item.is_file():
+                item.unlink()
+
+    response = client_with_session.get(ROUTE)
```
GibranAlfa: same as before, not calling the get changelog

tests/test_v1/test_project.py
```diff
@@ -296,6 +296,52 @@ async def test_get_project_already_in_session(
...
+    assert response.status_code == status.HTTP_404_NOT_FOUND
+
+
+async def test_get_changelog_permission_denied(
+    client_with_session: TestClient,
+    session_tmp_path: Path,
+    no_permissions: Callable[[str | Path], AbstractContextManager[None]],
+) -> None:
+    """Test 403 is returned when changelog cannot be read due to permissions."""
+    changelog_path = session_tmp_path / ".fmu"
```
GibranAlfa: this is not the changelog path

tests/test_services/test_changelog_service.py
```diff
@@ -0,0 +1,30 @@
...
+def test_get_changelog_contains_changeinfo_entries(
+    fmu_dir: ProjectFMUDirectory,
+) -> None:
+    """Test get_changelog returns entries of type ChangeInfo."""
+    service = ChangelogService(fmu_dir)
+
+    result = service.get_changelog()
+
+    for entry in result.root:
+        assert isinstance(entry, ChangeInfo)
```
GibranAlfa: these tests will pass if Log is empty, so a better way to do this is:
1. create a changelog entry:
```python
service = ChangelogService(fmu_dir)
fmu_dir.changelog.add_log_entry(
        ChangeInfo(
           ...
        )
    )
```
2. call get_changelog as you have done
```python
result = service.get_changelog()
```
3. assert `len(result) == 1`
4. assert concrete field in ChangeInfo, e.g. `assert result[0].change_type == ChangeType.update`, depending on what you add in the log entry

slangeveld (commented):
Looks like the test cases you have created covers the different scenarioas well. 

Still some work to do on the implementation of the tests, but Gibrans suggested changes should do the trick.

Take your time to understand what the tests are doing and why the different features like `patch` and `TestClient` are being used. Understanding of this will make it easier to create new tests later. Don't hesitate to ask if anything is unclear.

PR #325, opened 2026-03-20
ENH: Add RMS session handling + make user session lifetime 1 year

### LINKED ISSUE #235: Transferring data from old session to new does not work
In the endpoint to create a new session `POST /session` there is logic to transfer relevant data from an existing session to a new one.
This does not work and no data is transferred to the new session. 

The reason is that the `SessionManager.get_session` method will trigger a destroyal of the old session and raises a `SessionNotFoundError` if a session has expired.
https://github.com/equinor/fmu-settings-api/blob/b419ce8bf5df530528f9d28f913d8152b67605aa/src/fmu_settings_api/session.py#L201-L203

Hence the exception here is suppressed and the `old_session` variable is never defined and stays `None`. The `old_session` value is used in turn to check if data should be transferred.

https://github.com/equinor/fmu-settings-api/blob/b419ce8bf5df530528f9d28f913d8152b67605aa/src/fmu_settings_api/v1/routes/session.py#L70-L74

Existing tests does not pick up on it since we are testing to transfer data between a _non-expired_ session to a new one, hence we do not trigger the `SessionNotFoundError`.

slangeveld:
Resolves #235

(Update: Check the latest comment in this PR for the most accurate explanation to what has been changed)

Add logic for refreshing sessions and handling RMS sessions separately.

The session logic has been refactored to only accept refreshing of sessions through the new session refresh endpoint and not through internal logic in `get_session`. The RMS session has been updated with a field `expires_at` which allows for handling the RMS session and user session separately.

In addition the logic for destroying a session is moved to a new dependency DestroySessionIfExpiredDep that is injected into the relevant places.

Both the RMS session and user session are treated as sessions that can expire and be refreshed. At the moment the user session expiration time is set to 1 year so in practice we will not need that there, but it gives the flexibility to reduce this later if needed.

As the session logic affect a big part of the code (especially tests), the PR is quite large (mostly tests) so have tried to add comments for explanation and context at the relevant places. Let me know if anything is unclear.

**The new flow for creating, refreshing and destroying sessions will be:**
1. User creates a user session with a long expiration time with the `POST` endpoint
2. User opens an RMS project in the GUI for which it will be added to the session with a separat expiration time
3. Each time a call is made to any of the `fmu-settings-api` endpoints that requires a session, the DestroySessionIfExpiredDep is injected, making sure any expired RMS sessions are removed or user sessions are destroyed. 
4. The GUI needs to keep track of the session expiration time and make sure to refresh the RMS or user session before it has expired. If expired, the RMS session will be removed or the user session will be destroyed in the next request to `fmu-settings-api`.
5. No session state is transferred to new sessions. Either an existing session is refreshed (keeping the state as-is), an expired session is destroyed/removed or a new session is created. 

## Checklist

### DIFF
--- a/src/fmu_settings_api/session.py
+++ b/src/fmu_settings_api/session.py
@@ -272,9 +272,9 @@ async def destroy_fmu_session_if_expired(session_id: str) -> None:
     except SessionNotFoundError:
         return
     now = datetime.now(UTC)
-    rms_session_expiration = await get_rms_session_expiration(session.id)
+    rms_session_expiration = get_rms_session_expiration(session)
     if rms_session_expiration is not None and rms_session_expiration < now:
-        await remove_rms_project_from_session(session.id)
+        await remove_rms_project_from_session(session_id)
 
     if session.expires_at < now:
         await session_manager.destroy_session(session_id)
@@ -536,13 +536,12 @@ async def remove_rms_project_from_session(session_id: str) -> ProjectSession:
     return session
 
 
-async def get_rms_session_expiration(session_id: str) -> datetime | None:
+def get_rms_session_expiration(session: Session | ProjectSession) -> datetime | None:
     """Get the expiration time of an RMS session.
 
-    If the user session with session_id contains an open RMS session,
+    If the session contains an open RMS session,
     the expiration time of this RMS session is returned.
     """
-    session = await get_fmu_session(session_id)
     if isinstance(session, ProjectSession) and session.rms_session is not None:
         return session.rms_session.expires_at
     return None
--- a/src/fmu_settings_api/v1/routes/session.py
+++ b/src/fmu_settings_api/v1/routes/session.py
@@ -1,7 +1,6 @@
 """The main router for /api/v1/session."""
 
 import contextlib
-from pathlib import Path
 from textwrap import dedent
 from typing import Annotated
 
@@ -98,7 +97,7 @@ async def post_session(
                 id=session.id,
                 created_at=session.created_at,
                 expires_at=session.expires_at,
-                rms_expires_at=await get_rms_session_expiration(session.id),
+                rms_expires_at=get_rms_session_expiration(session),
                 last_accessed=session.last_accessed,
             )
         except SessionNotFoundError:
@@ -115,8 +114,7 @@ async def post_session(
     )
 
     with contextlib.suppress(FileNotFoundError, LockError):
-        path = Path.cwd()
-        project_fmu_dir = find_nearest_fmu_directory(path)
+        project_fmu_dir = find_nearest_fmu_directory()
         await add_fmu_project_to_session(session_id, project_fmu_dir)
 
     session = await get_fmu_session(session_id)
@@ -191,7 +189,7 @@ async def get_session(
     session_service: SessionServiceDep,
 ) -> SessionResponse:
     """Returns the current session in a serialisable format."""
-    return await session_service.get_session_response()
+    return session_service.get_session_response()
 
 
 @router.post(
--- a/tests/test_session_manager.py
+++ b/tests/test_session_manager.py
@@ -1229,50 +1229,29 @@ async def test_remove_fmu_project_from_session_closes_rms_project(
 async def test_get_rms_session_expiration() -> None:
     """Tests getting the RMS session expiration time from a session."""
     mocked_session = AsyncMock(spec=ProjectSession)
-    mocked_session_id = "mocked_id"
-    mocked_session.id = mocked_session_id
     mocked_rms_session = MagicMock()
     mocked_rms_session.expires_at = datetime.now(UTC)
     mocked_session.rms_session = mocked_rms_session
 
-    with patch("fmu_settings_api.session.session_manager") as mocked_session_manager:
-        mocked_get_session = AsyncMock(return_value=mocked_session)
-        mocked_session_manager.get_session.side_effect = mocked_get_session
+    rms_session_expiration = get_rms_session_expiration(mocked_session)
 
-        rms_session_expiration = await get_rms_session_expiration(mocked_session_id)
-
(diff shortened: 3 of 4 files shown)


src/fmu_settings_api/session.py
```diff
@@ -166,23 +168,21 @@ async def create_session(
...
             access_tokens=AccessTokens(),
         )
         await self._store_session(session_id, session)
 
         return session_id
 
-    async def get_session(
-        self: Self, session_id: str, extend_expiration: bool = True
-    ) -> Session | ProjectSession:
+    async def get_session(self: Self, session_id: str) -> Session | ProjectSession:
```
slangeveld: The `get_session` class method has been updated to only do what you would expect from its method name: getting the session from the session storage. The logic for extending and destroying the session has been moved out from here.

src/fmu_settings_api/models/session.py
```diff
@@ -17,5 +17,8 @@ class SessionResponse(BaseResponseModel):
     expires_at: datetime
     """Timestamp when the session will expire."""
 
+    rms_expires_at: datetime | None
```
slangeveld: This is what will keep track of when the RMS session has expired.

src/fmu_settings_api/v1/routes/session.py
```diff
@@ -81,17 +81,36 @@ async def post_session(
...
+    """Creates a new user session or adds a project to the existing one."""
     if fmu_settings_session:
-        with contextlib.suppress(Exception):
-            old_session = await session_manager.get_session(
-                fmu_settings_session, extend_expiration=False
-            )
+        try:
+            existing_session = await get_fmu_session(fmu_settings_session)
+            if isinstance(existing_session, ProjectSession):
+                response.set_cookie(
```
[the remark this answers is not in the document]
GibranAlfa: we don't need to set the cookie for an existing ProjectSession. The client sends credentials automatically and doesn’t read the session cookie or response id itself, so in this branch it already has the valid cookie it needs.

suggest to refactor this endpoint for readability:
```py
    session_id: str | None = None

    if fmu_settings_session:
        try:
            existing_session = await get_fmu_session(fmu_settings_session)
        except SessionNotFoundError:
            pass
        else:
            if isinstance(existing_session, ProjectSession):
                return SessionResponse(
                    id=existing_session.id,
                    created_at=existing_session.created_at,
                    expires_at=existing_session.expires_at,
                    rms_expires_at=await get_rms_session_expiration(
                        existing_session.id
                    ),
                    last_accessed=existing_session.last_accessed,
                )
            session_id = existing_session.id

    if session_id is None:
        session_id = await create_fmu_session(user_fmu_dir)
        response.set_cookie(
            key=settings.SESSION_COOKIE_KEY,
            value=session_id,
            httponly=True,
            secure=False,
            samesite="lax",
        )

    with contextlib.suppress(FileNotFoundError, LockError):
        path = Path.cwd()
        project_fmu_dir = find_nearest_fmu_directory(path)
        await add_fmu_project_to_session(session_id, project_fmu_dir)

    session = await get_fmu_session(session_id)

    return SessionResponse(
        id=session.id,
        created_at=session.created_at,
        expires_at=session.expires_at,
        rms_expires_at=None,
        last_accessed=session.last_accessed,
    )
```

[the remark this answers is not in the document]
tnatt: It feels like this could be simplified even more now when a session in theory doesn't expire..
Could we make this into a pure create session endpoint?
 
Currently, the GUI uses this endpoint to create an initial session and also calls it again whenever the API returns a "session expired" response. However, with the changes introduced this PR to extend a session lifetime this automatic session creation can be removed from the GUI - then there will be only one call to this endpoint. 

The dependency added will destroy an existing session if it has expired, could this endpoint if run multiple times raise SessionAlreadyExistError, or something simple? 

We already have the `get_session` endpoint for getting a session response for an existing session. and we have another endpoint for adding a project to the session. So we don't really need to run this endpoint more than the initial creation of the session.


```python
    if fmu_settings_session:
        try: 
            existing_session = await get_fmu_session(fmu_settings_session)
            return Message("Session already exists")
        except SessionNotFoundError:
            pass

    session_id = await create_fmu_session(user_fmu_dir)
    
    with contextlib.suppress(FileNotFoundError, LockError):
        path = Path.cwd()
        project_fmu_dir = find_nearest_fmu_directory(path)
        await add_fmu_project_to_session(session_id, project_fmu_dir)

    session = await session_manager.get_session(session_id)
    return SessionResponse(
        id=session.id,
        created_at=session.created_at,
        expires_at=session.expires_at,
        rms_expires_at=None
        last_accessed=session.last_accessed,
    )
```

src/fmu_settings_api/session.py
```diff
@@ -166,23 +168,49 @@ async def create_session(
...
     ) -> Session | ProjectSession:
+        """Renews an existing session by rotating its id and expiration.
+
+        Returns:
+            The renewed session instance stored under its new session id
+
+        Raises:
+            SessionNotFoundError: If the session does not exist
+        """
+        session = await self._retrieve_session(session_id)
```
slangeveld: I think we should use the wrapper function `get_fmu_session` here as it does exactly this (tries to get the session and returns SessionNotFoundError if it cant find it). The we dont need to update `last_accessed` on line 207 either.

tests/test_session_manager.py
```diff
@@ -45,23 +52,224 @@ def mock_rms_project() -> MagicMock:
     return MagicMock(close=MagicMock())
 
 
-def test_session_manager_init() -> None:
-    """Tests initialization of the SessionManager."""
-    assert session_manager.storage == SessionManager().storage == {}
+class TestSessionManagerClass:
```
slangeveld: There should be some tests for `renew_session()` here as well.

PR #344, opened 2026-04-08
ENH: Separate user and project restore-from-memory flows

### LINKED ISSUE #336: Separation of concern to restore user and project .fmu from memory
The current restore from memory endpoint automatically restore both user and project .fmu. With the latest `fmu-settings`, we can now call restore to only restore user or project .fmu. Thus, we need to distinguish them, maybe with separate endpoint or special handling in the service logic.

We can also list the restorable files, and if there's no restorable files and we don't need to call restore.

GibranAlfa:
Resolves #336 

Feedback from joar for this [PR](https://github.com/equinor/fmu-settings-gui/pull/269) in GUI:
1. First comment (ref [here](https://github.com/equinor/fmu-settings-gui/pull/269#discussion_r2918795068)): We need to check if there's any missing/deleted files before restoring, so we don't need to restore if there's no missing files, and we can tell the users what files are restored if there's any.
2. Second comment (ref [here](https://github.com/equinor/fmu-settings-gui/pull/269#discussion_r2918970887)): It's better to split the restore flow for user .fmu and project .fmu files, so the project page only restore project .fmu files and we can have a separate page to restore user .fmu files.

This PR added two endpoints (`GET /restore/check` and `POST /restore`) to both project route and session route to check restorable files and to restore missing files.

## Checklist

### DIFF
--- a/src/fmu_settings_api/deps/project.py
+++ b/src/fmu_settings_api/deps/project.py
@@ -20,7 +20,7 @@ async def get_project_service(
 ProjectServiceDep = Annotated[ProjectService, Depends(get_project_service)]
 
 
-async def get_restore_project_service(
+async def get_project_service_for_restore(
     session: SessionDep,
 ) -> ProjectService:
     """Returns a ProjectService for restore routes, even if .fmu is deleted."""
@@ -34,5 +34,5 @@ async def get_restore_project_service(
 
 
 ProjectServiceForRestoreDep = Annotated[
-    ProjectService, Depends(get_restore_project_service)
+    ProjectService, Depends(get_project_service_for_restore)
 ]
--- a/src/fmu_settings_api/deps/project.py
+++ b/src/fmu_settings_api/deps/project.py
@@ -33,6 +33,6 @@ async def get_restore_project_service(
     return ProjectService(session.project_fmu_directory)
 
 
-ProjectRestoreServiceDep = Annotated[
+ProjectServiceForRestoreDep = Annotated[
     ProjectService, Depends(get_restore_project_service)
 ]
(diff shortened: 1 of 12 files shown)


src/fmu_settings_api/deps/project.py
```diff
@@ -17,3 +18,21 @@ async def get_project_service(
...
+            status_code=401,
+            detail="No FMU project directory open",
+        )
+
+    return ProjectService(session.project_fmu_directory)
+
+
+ProjectRestoreServiceDep = Annotated[
+    ProjectService, Depends(get_restore_project_service)
+]
```
GibranAlfa: need to add a new deps instead of using the current `ProjectServiceDep` since it requires `project_fmu_directory.path` to already exist, which defeats the purpose of restoring missing project .fmu files.

mferrera (approved):

src/fmu_settings_api/deps/project.py
slangeveld: This is a bit misleading as there is no service called ProjectRestoreService. I think it is better to use the ProjectServiceDep here and handle this case in the ProjectService constructor

GibranAlfa: The problem is not about the `ProjectService` but the `ProjectServiceDep` requires that the project .fmu exists https://github.com/equinor/fmu-settings-api/blob/f5ab9ca7edda49e7edc2de626fb11677274b0df0/src/fmu_settings_api/deps/session.py#L70-L75
so I need another dep that doesn't require project .fmu. Modifying the current `ProjectServiceDep` to cater to this restore endpoint need is also not a good idea because it makes sense to check project .fmu exists before calling methods in `ProjectService`.

I agree that name is a bit misleading, so I will rename it to `ProjectServiceForRestoreDep `

slangeveld: Fair enough. Renaming the dependency helps for readability. Maybe also rename the function `get_restore_project_service` accordingly then

Issue #200, opened 2025-11-24
SMDA masterdata search to only support searching for a single field


joargr:
Today the SMDA masterdata search endpoint (`POST /api/v1/smda/masterdata`) supports searching for multiple fields, as the search parameter can be a list of multiple fields. The results that come back from the SMDA API is combined by FMU Settings API into a single result, merging all data into one result set.

The data from the SMDA API has a key `field_coordinate_system`, where the value is the default coordinate system for a field. When results from multiple fields are merged into one results, it becomes difficult or impossible to associate the default coordinate system value to a specific field. The implementation of issue #175 tries to deal with this, by moving default coordinate systems for one or more fields to the beginning of the list of all coordinate systems, but when the search results from the FMU Settings API endpoint is processed it is not possible to always get correct data when the search has been for multiple fields. Ie. two fields could have two different default coordinate systems, or they could share the same coordinate system. When processing the results from FMU Settings API it is not possible to know whether the first one or the first two coordinate systems in the list are defaults.

As of now, the FMU Settings GUI only specifies a single field in a single search. When it is necessary to get data on multiple fields, the GUI executes multiple searches, with one field per search. For each of these search results, the value from the `field_coordinate_system` is read, and used in the GUI. The updated logic in the API search result handling is not necessary, as finding the default coordinate system is done by reading the `field_coordinate_system` value, and not relying on the coordinate system being at the beginning of the list of all coordinate systems.

Sorting the list of all coordinate systems and determining how it should be displayed to the user should be done in the GUI. The API should thus not need to do any sorting itself, especially since the results already include necessary information about default values (the `field_coordinate_system` value) - and this field can be relied on since the GUI knows that the results are for only one field.

It is suggested that the SMDA masterdata search endpoint only support searching for a single field at a time, and that the logic processing the search results from the SMDA API be simplified (effectively reverting the changes done when implementing issue #175). When searching for only one field is supported, the results from the SMDA API can be served to the client with little need for processing.

PR #345, opened 2026-04-09
ENH: Add UUID-based SMDA lookup and country to field search

### LINKED ISSUE #323: Update endpoint for field search to return country
Update endpoint for field search must also return country for the field.

GibranAlfa:
Resolves #323 

- Add country to the SMDA field search response (in addition to identifier and uuid).
- Masterdata lookup now accepts a selected field model with an optional uuid, and uses that uuid for exact SMDA lookup when it is provided.

  This is added to solve that Raia field problem. To disambiguate fields that exist in multiple countries, we can search for both the field identifier and country, but it's even better to just use the uuid because it's unique. The idea is for the GUI to display field and country, so the user can select which field they want to do masterdata lookup, but the GUI actually sends the field identifier and the field uuid to the API to search for the correct masterdata.

- The changes in this PR is adapting to the fact that GUI is only sending one field when doing masterdata search (described in this issue #200) so we assume there's only one UUID sent to the POST smda/masterdata and there will be only one result from the search. Separate PR should be made to solve that issue and clarify that the API only support one field instead of list of field (this change would also break the GUI so not desired for now).

## Checklist

### DIFF
--- a/src/fmu_settings_api/interfaces/smda_api.py
+++ b/src/fmu_settings_api/interfaces/smda_api.py
@@ -2,6 +2,7 @@
 
 from collections.abc import Sequence
 from typing import Any, Final
+from uuid import UUID
 
 import httpx
 
@@ -72,14 +73,23 @@ class SmdaAPI:
         return res.status_code == httpx.codes.OK
 
     async def field(
-        self, field_identifiers: Sequence[str], columns: Sequence[str] | None = None
+        self,
+        field_identifiers: Sequence[str] | None = None,
+        field_uuid: UUID | None = None,
+        columns: Sequence[str] | None = None,
     ) -> httpx.Response:
         """Searches for a field identifier in SMDA."""
         _projection = "identifier,uuid" if columns is None else ",".join(columns)
+        json: dict[str, Any] = {"_projection": _projection}
+
+        if field_identifiers:
+            json["identifier"] = field_identifiers
+        if field_uuid is not None:
+            json["uuid"] = [str(field_uuid)]
 
         return await self.post(
             SmdaRoutes.FIELDS_SEARCH,
-            json={"_projection": _projection, "identifier": field_identifiers},
+            json=json,
         )
 
     async def country(
--- a/src/fmu_settings_api/services/smda.py
+++ b/src/fmu_settings_api/services/smda.py
@@ -88,9 +88,13 @@ class SmdaService:
 
         if not field_results:
             requested_identifiers = [field.identifier for field in smda_fields]
-            raise ValueError(
-                f"No fields found for identifiers: {requested_identifiers}"
-            )
+            requested_uuids = [
+                str(field.uuid) for field in smda_fields if field.uuid is not None
+            ]
+            error_msg = f"No fields found for identifiers: {requested_identifiers}"
+            if requested_uuids:
+                error_msg += f" and UUIDs: {requested_uuids}"
+            raise ValueError(error_msg)
 
         field_items = [FieldItem.model_validate(field) for field in field_results]
         field_identifiers = [field.identifier for field in field_items]
--- a/src/fmu_settings_api/services/smda.py
+++ b/src/fmu_settings_api/services/smda.py
@@ -16,6 +16,7 @@ from fmu_settings_api.models.smda import (
     SmdaField,
     SmdaFieldSearchResult,
     SmdaMasterdataResult,
+    SmdaSelectedField,
     SmdaStratigraphicUnitsResult,
     StratigraphicUnit,
 )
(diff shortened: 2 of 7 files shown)


src/fmu_settings_api/interfaces/smda_api.py
```diff
@@ -72,14 +73,23 @@ async def health(self) -> bool:
         return res.status_code == httpx.codes.OK
 
     async def field(
-        self, field_identifiers: Sequence[str], columns: Sequence[str] | None = None
+        self,
+        field_identifiers: Sequence[str] | None = None,
+        field_uuid: UUID | None = None,
```
GibranAlfa, line 78: this field search now support optional uuid and field_identifiers. a bit weird to see both are optional, but this is made to support POST smda/field that only use field_identifiers and POST smda/masterdata that support both of them (but only use uuid on the logic if uuid is present). as this is only an internal method and all the callers correctly set them i think this is fine.

src/fmu_settings_api/services/smda.py
```diff
@@ -35,34 +36,60 @@ async def check_health(self) -> bool:
...
+            field_res = await self._smda.field(
+                field_uuid=smda_fields[0].uuid,
+                columns=[
+                    "country_identifier",
+                    "identifier",
+                    "projected_coordinate_system",
+                    "uuid",
+                ],
+            )
+            field_results = field_res.json()["data"]["results"]
```
GibranAlfa, line 71: if uuid is provided, we use only uuid to do the field search for masterdata lookup, otherwise use the field identifier like before.

joargr:
There is an edge case which doesn't seem to be handled well. Some fields are associated with two countries, and when searching for such a field the returned result shows only one country for the field. An example of this is the field FRIGG, which is associated with the countries Norway and United Kingdom. As of now the search result shows Norway, with no reference to United Kingdom. The Opus SMDA viewer shows both countries, so I assume that the SMDA API also gives this information.

As the `country` field in the search result is merely an informational field, suggest that the value shown is a comma separated list of countries. For most of the fields this value will only show one country, but for those few fields which are associated with more than one country, the `country` field would then show something like "Norway, United Kingdom".

Related to a field search for FRIGG, I see that a masterdata search (`POST /api/v1/smda/masterdata`) for this field only lists Norway in the `country` list. Is this a bug? Or does the SMDA API only return one country for this field? If the SMDA API does indeed return two countries, then the Settings API should also return both countries.

src/fmu_settings_api/interfaces/smda_api.py
joargr: When a masterdata search (`POST /api/v1/smda/masterdata`) is performed and the search criteria is specified with both `identifier` and `uuid` values, if there is no match the error message will only list the `identifier` value. If both values are specified as the search criteria then the error message should also show this.

src/fmu_settings_api/v1/routes/smda/main.py
```diff
@@ -145,9 +146,10 @@ async def get_health(smda_service: SmdaServiceDep) -> Ok:
         A route to search SMDA for an field (asset) by its named identifier.
```
joargr: Typo.

```suggestion
        A route to search SMDA for a field (asset) by its named identifier.
```

GibranAlfa:
ok so this is an interesting case, with Frigg, SMDA API return Norway as the only country_identifier, but have Norway and UK in "country_list" attribute, which is not used by the current FMU Settings API.

I would assume that country_identifier which we use now is unique, so maybe in the eye of SMDA, Frigg is in Norway. Maybe the part of Frigg in UK is not an interest to Equinor as this field is operated by Aker BP and maybe that UK section is owned by different operator.

So I think current implementation that rely on country identifier is ok, unless there's an argument otherwise.

GibranAlfa:
after some more investigation:

SMDA explain country_identifer as "Name of the main country" and country_list as "List of countries where the field is located"

This is also the case for Statfjord field for example, with has Norway as country_identifier and Norway and UK in the country_list. So I think we should only should the country identifier as it is now, to make it clear which country this field mainly belongs to.

List of fields that exist in both Norway and UK:
image

joargr:
There seems to be a bug related to searching masterdata for multiple fields and specifying both the `identifier` and the `uuid` value as the search criteria. When searching for multiple fields, if the first of those fields are specified only with `identifier` then the result includes data for all fields. But if the first of those fields are specified with `uuid` value as well, then the result includes data only for the first field. In other words:

- Search for `[ (identifier: A), (identifier: B)]` returns data for both fields
- Search for `[ (identifier: A, uuid: 1), (identifier: B, uuid: 2) ]` returns only data for field A
- Search for `[ (identifier: A), (identifier: B, uuid: 2)]` returns data for both fields
- Search for `[ (identifier: A, uuid: 1), (identifier: B) ]` returns only data for field A

It seems that if the first field is specified with an `uuid` value then subsequent fields in the search criteria are ignored.

While FMU Settings GUI only searches for one field at a time (searches for multiple fields are done in parallel), this is a bug which should be fixed, as the API does support searching for multiple fields in one search, and the results should be consistent in how the search criteria is handled.

joargr:
> SMDA explain country_identifer as "Name of the main country" and country_list as "List of countries where the field is located"
> 
> This is also the case for Statfjord field for example, with has Norway as country_identifier and Norway and UK in the country_list. So I think we should only should the country identifier as it is now, to make it clear which country this field mainly belongs to.

Interesting research, good to have a list of such edge cases. I guess it's ok to only show and use the main country. The description of the endpoints should be updated to include this information, to say that only the main country is included in the results and not the list of countries.

But for those fields that are located in more than one country, is it necessary to include those other countries in the masterdata in the project? @tnatt do you know if this information is needed in RMS/ERT/other systems, and therefore needs to be added in the masterdata editing in the GUI?

GibranAlfa:
> There seems to be a bug related to searching masterdata for multiple fields and specifying both the `identifier` and the `uuid` value as the search criteria. When searching for multiple fields, if the first of those fields are specified only with `identifier` then the result includes data for all fields. But if the first of those fields are specified with `uuid` value as well, then the result includes data only for the first field. In other words:
> 
> * Search for `[ (identifier: A), (identifier: B)]` returns data for both fields
> * Search for `[ (identifier: A, uuid: 1), (identifier: B, uuid: 2) ]` returns only data for field A
> * Search for `[ (identifier: A), (identifier: B, uuid: 2)]` returns data for both fields
> * Search for `[ (identifier: A, uuid: 1), (identifier: B) ]` returns only data for field A
> 
> It seems that if the first field is specified with an `uuid` value then subsequent fields in the search criteria are ignored.
> 
> While FMU Settings GUI only searches for one field at a time (searches for multiple fields are done in parallel), this is a bug which should be fixed, as the API does support searching for multiple fields in one search, and the results should be consistent in how the search criteria is handled.

This is intentional as I mentioned in the PR desc here:
> - The changes in this PR is adapting to the fact that GUI is only sending one field when doing masterdata search (described in this issue #200) so we assume there's only one UUID sent to the POST smda/masterdata and there will be only one result from the search. Separate PR should be made to solve that issue and clarify that the API only support one field instead of list of field (this change would also break the GUI so not desired for now).

as I want to move the API to align with GUI behavior that only do one field search at a time (unless there's an idea to do multiple search from the GUI in one search instead of doing it in parallel). The idea is also to still support list for now (even tough only one uuid is used on the logic) so that it doesn't break the current GUI implementation.

tnatt:
> But for those fields that are located in more than one country, is it necessary to include those other countries in the masterdata in the project? @tnatt do you know if this information is needed in RMS/ERT/other systems, and therefore needs to be added in the masterdata editing in the GUI?

I can't think of anywhere that the information is needed, specifying the main country should be fine. If for some reason there is a need in the future, we can modify it then.

joargr:
> as I want to move the API to align with GUI behavior that only do one field search at a time (unless there's an idea to do multiple search from the GUI in one search instead of doing it in parallel). The idea is also to still support list for now (even tough only one uuid is used on the logic) so that it doesn't break the current GUI implementation.

Fair enough. But perhaps you can update the description of the endpoint with this, to describe that if a field UUID value is provided as well then the endpoint will only search for one field, and search for the first field if several fields are provided as the search criteria.

joargr (approved):
This looks good.

PR #361, opened 2026-04-23
ENH: Add horizon uuid in strat units response

### LINKED ISSUE #360: Add horizon uuid in the strat_units endpoint response
Because we want to map horizon from RMS to official sources (e.g. SMDA). We can do it just for the name like now, but having uuid will make it unique and not confusing.

GibranAlfa:
Resolves #360 

Added top and base horizon uuid to the strat units response by taking the top and base result from the strat units SMDA endpoint to search for it's UUID in strat-surface-names SMDA endpoint.

If horizon lookup fails, it will log the error, set the uuid to null, and not failing the strat_units endpoint.

Example response now from TROLL strat column:
```json
"stratigraphic_units": [
    {
      "identifier": "NORDLAND GP.",
      "uuid": "ad215072-12e7-2e4b-e053-c918a4881b5c",
      "strat_unit_type": "group",
      "strat_unit_level": 1,
      "top": "NORDLAND GP. Top",
      "top_uuid": "ad215071-9ca5-2e4b-e053-c918a4881b5c",
      "base": "NORDLAND GP. Base",
      "base_uuid": "ad215071-72ca-2e4b-e053-c918a4881b5c",
      "top_age": 0.001,
      "base_age": 16,
      "strat_unit_parent": null,
      "strat_column_type": "lithostratigraphy",
      "color_html": "#00bc48",
      "color_r": 0,
      "color_g": 188,
      "color_b": 72
    },
```

## Checklist


src/fmu_settings_api/interfaces/smda_api.py
```diff
@@ -20,6 +20,7 @@ class SmdaRoutes:
     STRAT_COLUMN_AREAS_SEARCH: Final[str] = "smda-api/strat-column-areas/search"
     STRAT_UNITS_SEARCH: Final[str] = "smda-api/strat-units/search"
     COORDINATE_SYSTEM_SEARCH: Final[str] = "smda-api/crs/search"
+    HORIZON_SEARCH: Final[str] = "smda-api/strat-surface-name-alias/search"
```
tnatt: Is this the correct endpoint to use here, or should it be the `smda-api/strat-surface-names/search` instead?

GibranAlfa: both are correct, i can change it to strat-surface-names instead for clarity. also "Sognefjord Fm. 6.2 TRO Top" still doesn't have uuid in SMDA, maybe because it's a horizon of a subzone. I guess this indicates that we should choose the coarsest horizon when choosing a horizon on a same level

GibranAlfa: just did the search again for "Sognefjord Fm. 6.2 TRO Top" and now they have uuid! 🤯 they just added them i guess, so still need to decide which horizon to pick

src/fmu_settings_api/models/smda.py
```diff
@@ -108,9 +108,15 @@ class StratigraphicUnit(BaseResponseModel):
     top: str = Field(examples=["VIKING GP. Top"])
     """The identifier (name) of the stratigraphic unit top pick (horizon)."""
 
+    top_horizon_uuid: UUID | None
```
tnatt: For consistency should we either rename the `top` and `base` fields to `top_horizon_identifier` and `base_horizon_identifier` ? 

we could also consider making the `top` an `base` an object. e.g:
```
top_horizon:
  uuid:
  identifier:
  age:
```

GibranAlfa: we could, but will need more changes in the API to adjust to the new top and base name, and probably the gui code as well if it's already used. also top and base are the results from SMDA API, so it's better to match the API model with the SMDA response.

src/fmu_settings_api/services/smda.py
```diff
@@ -270,3 +288,40 @@ async def _get_coordinate_systems(
...
+                continue
+
+            if isinstance(response, BaseException):
+                raise response
+
+            horizon_results = response.json()["data"]["results"]
+            if not horizon_results:
+                continue
+
+            horizon_uuid = horizon_results[0].get("uuid")
```
tnatt: I guess it is safe to always take the first item here since in theory there should not be more than 1 result for a given identifier, if not there are data quality issues in smda. But anyway, is the length of the results worth checking?

GibranAlfa: yeah, I thought about it, not sure what to do if length is more than 1, do we tell user or just swallow the error? because of that I just take the first result here

src/fmu_settings_api/models/smda.py
tnatt: Then I suggest to rename this to `top_uuid` and `base_uuid` to make it as equal to the rest of the response from SMDA API 👍

src/fmu_settings_api/services/smda.py
tnatt: I agree, not easy 🤔  .. I don't think the users would like to know about this, but it feels like we should record it somehow to report it back to smda.

src/fmu_settings_api/services/smda.py
```diff
@@ -190,9 +194,23 @@ async def get_stratigraphic_units(
                 f"{strat_column_identifier}"
             )
 
+        horizon_identifiers: set[str] = set()
```
tnatt: SMDA refers to `horizons` as `surfaces.` Since this code is for fetching smda data, I think we should rename all horizon references here to surface. Then we are consistent, since we already use the smda term `unit` and not the rms term `zone`

GibranAlfa: this is only internal variable in the fmu-settings-api to search for horizon, to search for the horizon uuid in that strat-surface-names, we send `identifier` as the attribute for horizon/surface name, but i will change it to surface as well if it's more clear

src/fmu_settings_api/services/smda.py
GibranAlfa: we can log it, but should also actively monitor user logging to know if this happens, or surface it to the user (which is unnecessary). I asked in smda slack channel if this can even happen, I suspect not.

tnatt: perfect! Lets see what they respond 👍

src/fmu_settings_api/services/smda.py
```diff
@@ -190,9 +194,23 @@ async def get_stratigraphic_units(
                 f"{strat_column_identifier}"
             )
 
+        surface_identifiers: set[str] = set()
+        for strat_unit_data in strat_unit_results:
+            for identifier in (strat_unit_data["top"], strat_unit_data["base"]):
+                if identifier:
+                    surface_identifiers.add(identifier)
+
+        horizon_uuids = await self._get_horizon_uuids(surface_identifiers)
```
tnatt: My comment above was referring to all the internal variable and function names 🙂. I know they are internal, but still would prefer all `horizon` references to be renamed to `surface`. E.g. for the example here, and all the code in the `_get_horizon_uuids` function
```suggestion
        surface_uuids = await self._get_surface_uuids(surface_identifiers)
```

tnatt (approved):
LGTM

src/fmu_settings_api/services/smda.py
GibranAlfa: SMDA confirmed in slack channel that the result would always be unique (only 1) so we don't need to handle >1 result

PR #368, opened 2026-04-29
ENH: Align mappings endpoint with the new data model

### LINKED ISSUE #362: Support unmappable relation type in mappings endpoint
As we introduced a new `relation_type = unmappable` and removed `relation_type = equivalent` in this issue (https://github.com/equinor/fmu-datamodels/issues/205), we should adjust the API logic due to that change.

1. Update the tests to not test `relation_type = equivalent`
2. Remove grouping logic as the GUI will handle it. Return raw mappings.json content

### LINKED ISSUE #363: Use group validation from mappings model
Instead of using `MappingGroup` as the group validator in the API, we should use the group validation from stratigraphy and wellbore mappings model instead and build response model from that. After this, we can move away from using `MappingGroup`

GibranAlfa:
Resolves #362 
Resolves #363 

- Removed `MappingGroup` from the face of the earth
- Removed equivalent and support unmappable case
- Adjusted logic to use the `InternalMappings`
- Clean up some private imports
- Had a discussion with Joar to align the mappings endpoint as `GET` and `PUT /project/mappings/{mapping_type}/{source_system}` so the GUI only call once for read and write
- Now get mappings endpoint return empty list when mappings.json file is not found

Can be tested now @joargr 

## Checklist


src/fmu_settings_api/v1/routes/project.py
```diff
@@ -1384,6 +1371,14 @@ async def put_mappings(
             detail=f"Invalid mappings: {'; '.join(errors)}",
         ) from e
     except ValueError as e:
+        if str(e).startswith("Invalid content in resource file for 'MappingsManager:"):
```
GibranAlfa: added this to return a readable error instead of pydantic validation error that is cryptic when we have invalid mappings.json already installed locally

tests/test_services/test_mappings_service.py
```diff
@@ -23,159 +25,204 @@ def mappings_service(fmu_dir: ProjectFMUDirectory) -> MappingsService:
     return MappingsService(fmu_dir)
```
slangeveld, line 26: There should also be a test for the happy case of `get_internal_mappings_by_source_system()` here. When it does not fail.

tests/test_v1/test_project.py
```diff
@@ -3245,21 +3283,38 @@ async def test_get_mappings_stratigraphy_value_error(
...
+        response = client_with_project_session.get(f"{ROUTE}/mappings/stratigraphy/rms")
     assert response.status_code == status.HTTP_400_BAD_REQUEST
     assert response.json() == {"detail": "Mapping type 'wells' is not yet supported"}
 
 
-async def test_get_mappings_stratigraphy_file_not_found(
+async def test_get_mappings_stratigraphy_invalid_existing_file_value_error(
+    client_with_project_session: TestClient,
+) -> None:
+    """Test 400 returns a stable message for invalid stored mappings."""
```
slangeveld: This docstring is a bit confusing and could probably be written a bit more clear

tests/test_v1/test_project.py
```diff
@@ -3296,26 +3350,26 @@ async def test_put_mappings_stratigraphy_success(
...
     )
     assert response.status_code == status.HTTP_200_OK
     response_data = response.json()
     assert "message" in response_data
     assert "stratigraphy" in response_data["message"]
-    assert "rms" in response_data["message"]
-    assert "smda" in response_data["message"]
 
 
 async def test_put_mappings_stratigraphy_creates_file_if_missing(
```
slangeveld: This is already tested in `test_update_internal_mappings_by_source_system_creates_mappings_file_if_missing()`.
The tests here testing the mappings endpoints should not need to test the logic of the mappings_service methods. It should only need to test the logic in the endpoint itself and just assert that the mapping_service methods are called.

This is to separate concerns and not add unnecessary depedencies between the endpoint and service layers. A change in the service layer should only need to trigger changes for the tests of the service layers, not the endpoint layer.

slangeveld (approved):
Nice and clean 🚀 

A few suggested changes for the structuring of the tests. In general we should aim to only test the logic for the relevant layer we are testing. For example:
* Tests for the mapping_service layer should test the logic in the mapping_service methods
* Tests for the mapping endpoint should only care about testing the logic in the mapping endpoint itself. What is returned from the mapping_service layer should not be a concern for the endpoint layer and hence not tested.

joargr (approved):
Looks good! Initial use shows that the GUI can use the updated endpoints as intended.

PR #374, opened 2026-05-13
ENH: Add interface to interact with `WellboreMappings` files

### LINKED ISSUE #373: Add interface to interact with `WellboreMappings` files
We need an interface and services that interact with `WellboreMappings` files. It should be able to:

1. read `rms_eclipse.csv` that has the rms to simulator mappings and convert it to the well mappings model
2. write from well mappings model to `rms_eclipse.csv`
3. write from well mappings model to `rms_eclipse.renaming_table`
4. write from well mappings model to `rms_pdm.renaming_table`

GibranAlfa:
Resolves #373

Added `WellboreMappingsFileIO` interface and import export methods in `MappingsService`. Endpoints will be added in separate PR.

## Checklist


src/fmu_settings_api/services/mappings.py
```diff
@@ -41,6 +45,45 @@ def update_internal_stratigraphy_mappings(
...
+        self: Self, relative_path: str | Path | None = None
+    ) -> InternalWellboreMappings:
+        """Import RMS-to-simulator wellbore mappings from an rms_eclipse CSV file."""
+        self._fmu_dir._lock.ensure_can_write()
+        wellbore_mappings = self._wellbore_mappings_file_io.read_rms_eclipse_csv(
+            relative_path
+        )
+        return self._fmu_dir.mappings.update_internal_wellbore_mappings(
+            wellbore_mappings
+        )
```
GibranAlfa, line 72: Imported mappings from rms_eclipse.csv will be automatically saved to mappings.json in .fmu. I think we want to store it immediately and then users can make adjustment in the GUI. I think this is the easiest implementation-wise in the GUI, let user import current mappings first, and then continue mapping from that.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,330 @@
...
+                        target_uuid=None,
+                    )
+                )
+
+        return InternalWellboreMappings(root=mappings)
+
+    def write_rms_eclipse_csv(
+        self: Self,
+        wellbore_mappings: InternalWellboreMappings,
+        relative_path: str | Path | None = None,
```
slangeveld: Still not sure if we should allow for a default path. Especially for the writing operations. It should be very clear to the user where they are actually writing to.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,330 @@
...
+    ) -> None:
+        """Write wellbore mappings to a tab-separated renaming table."""
+        renaming_table_path.parent.mkdir(parents=True, exist_ok=True)
+
+        with renaming_table_path.open("w", encoding="utf-8", newline="") as file_handle:
+            file_handle.write(f"{header}\n")
+            for mapping in wellbore_mappings:
+                file_handle.write(f"{mapping.source_id}\t{mapping.target_id}\n")
+
+    def _filter_wellbore_mappings(
```
slangeveld: I am wondering if this method should live in the mappings service instead. The service should contain all the business logic. The interface should only do the reading and writing to file. As it is defined now I would expect the `write_rms_eclipse_csv` method to take the mappings it gets as input and write it to a csv. Now it also does some filtering as well. I think it would be more clean to do that filtering in the service first and only have logic for writing the mappings to csv in this interface.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,330 @@
...
+            )
+
+        self._write_wellbore_renaming_table(
+            filtered_wellbore_mappings,
+            renaming_table_path=renaming_table_path,
+            header="SETNAMES pdm\trms",
+        )
+        return resolved_relative_path
+
+    def _resolve_path(
```
slangeveld: Can this method be moved to the mappings service? Looks like all the public methods in this interface call this internal method as the first thing they do so maybe the path can already be resolved in the mappings service an given as input to the public methods in this interface?

slangeveld (commented):
Very nice to get do these read and write operations into a separate interface. I think we can still separate concern here a bit more, so that the methods in this interface only takes care about reading and writing from file (see comments).

In general we aim to make the service layer such that is it very clear from reading the methods in there what is happening. We can create interfaces and helper functions to hide away logic, but what is left in the service layer should still be a clear set of instructions telling the full story of what is happening.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
GibranAlfa: I still prefer having a default path, with clear messaging in the GUI of what the default path is. Maybe @tnatt can weigh in here on user perspective.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
GibranAlfa: yes, we can move this out to the service layer 👍

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
GibranAlfa: i would keep this in the interface. path resolution and validation is part of the file access concern, and keeping it there keeps the safety check close to the actual read/write operation instead of spreading that logic into the service layer.

tests/test_services/test_mappings_service.py
```diff
@@ -279,3 +279,141 @@ def test_update_internal_mappings_by_source_system_rejects_mismatched_source_sys
...
+            return_value=expected_path,
+        ) as write_mock,
+    ):
+        mappings_mock.return_value = stored_mappings
+        assert (
+            mappings_service.export_pdm_rms_renaming_table(expected_path)
+            == expected_path
+        )
+
+    write_mock.assert_called_once_with(filtered_mappings, expected_path)
```
slangeveld: Think it would be good to also have small test of the `_filter_wellbore_mappings` method here

slangeveld (approved):
Looks good to me 👍  Let's just wait for Thereses opinion on the default paths

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
tnatt: I see your point Sindre, but as long as we make it clear from the GUI side - I don't have a strong opinion about it!

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,278 @@
...
+        self: Self, relative_path: str | Path | None, default_path: Path
+    ) -> tuple[Path, Path]:
+        """Resolve an optional project-relative path and return both path forms."""
+        resolved_path = self._fmu_dir.resolve_path_inside_project(
+            Path(relative_path or default_path)
+        )
+
+        return resolved_path, resolved_path.relative_to(self._fmu_dir.base_path)
```
tnatt: While I can understand the need for checking that a path resolves inside the project, I still don't understand the need for returning both the resolved and relative path here. 

Could we not simply make due with the full resolved path? 

```suggestion
        return self._fmu_dir.resolve_path_inside_project(
            Path(relative_path or default_path)
        )
```

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,278 @@
...
+                represented in the pdm_rms.renaming_table format.
+        """
+        renaming_table_path, resolved_relative_path = self._resolve_path(
+            relative_path, self.PDM_RMS_RENAMING_TABLE_PATH
+        )
+        if not wellbore_mappings:
+            raise ValueError(
+                "No wellbore mappings available to write to pdm_rms.renaming_table"
+            )
+
```
tnatt, line 186: Now we are relying on that  the mappings are filtered correctly before entering here, but should we not have a check that it actually is? E.g. checking that all mappings have the expected `source_system` and `target_system` inside this function before proceeding.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,278 @@
...
+        return resolved_path, resolved_path.relative_to(self._fmu_dir.base_path)
+
+    def _write_wellbore_renaming_table(
+        self: Self,
+        wellbore_mappings: list[InternalWellboreIdentifierMapping],
+        *,
+        renaming_table_path: Path,
+        header: str,
+    ) -> None:
+        """Write wellbore mappings to a tab-separated renaming table."""
```
tnatt: I suggest having the `source_system` and `target_system` as input arguments instead of the `header` on format `header="SETNAMES pdm\trms"`. This way we would not mess up the order of the column names as easily, and we could even default to take the `target_system` and `source_system` from the mappings. 
WDYT ? 🙂 

```suggestion
        source_system: str | None = None,
        target_system: str  | None = None,
    ) -> None:
        """Write wellbore mappings to a tab-separated renaming table."""   
        source_system = source_system or wellbore_mappings[0].source_system
        target_system = target_system or wellbore_mappings[0].target_system
        header = f"SETNAMES {source_system}\t{target_system}"
```

src/fmu_settings_api/services/mappings.py
```diff
@@ -41,6 +47,84 @@ def update_internal_stratigraphy_mappings(
...
+
+    def export_pdm_rms_renaming_table(
+        self: Self, relative_path: str | Path | None = None
+    ) -> Path:
+        """Export PDM-to-RMS wellbore mappings as pdm_rms renaming table."""
+        self._fmu_dir._lock.ensure_can_write()
+        filtered_wellbore_mappings = self._filter_wellbore_mappings(
+            wellbore_mappings=self._fmu_dir.mappings.internal_wellbore_mappings,
+            source_system=DataSystem.pdm,
+            target_system=DataSystem.rms,
```
tnatt: BTW why do we have `pdm` as the `source_system` here, should that be the opposite way? For stratigraphy mappings we have `rms` as the `source_system` right?

tnatt (commented):
In general this looks nice, and well structured 🙂
I have a couple of concerns and some comments.

My main concern is that we probably should rename all `eclipse` references to `simulator`.
Even though users have eclipse in there files now, I don't think we should forward that reference as our default. A lot of the users are now on OPM Flow. 
So I think we should default everything to follow the simulator naming, especially for writing operations. We could potentially allow users to change the default header names for writing - but not sure if that is necessary actually as changing this inside rms is a trivial thing - and all tools that uses the rms_eclipse.csv (that I know of) allows users to specify the name of the csv file as well the column names.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
GibranAlfa: so the relative path was intended to be part of the return messages in the endpoints, but I just remembered that we actually don't return path anymore to prevent users looking into it, so I will remove the relative path altogether from write methods

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
GibranAlfa: i think this is about separation of concern. this interface is now intended to just read and write wellbore mappings, the caller (which is the mapping service) is the one deciding what mappings to be exported, and that's why we have the filter in the service now. The mapping service has to make sure it filtered correctly, and I may add a check after filtering in the service before calling the interface.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
GibranAlfa: (from planning meeting discussion): we decided to just use source and target system as the header, not following the current standard

src/fmu_settings_api/services/mappings.py
GibranAlfa: so I don't see any existing rms_pdm.csv or rms_pdm.renaming_table in Drogon, but there's pdm_rms.renaming_table https://github.com/equinor/fmu-drogon/blob/main/rms/input/well_modelling/well_info/pdm_rms.renaming_table , so this is added to support that, i'm not sure if we need it or not now

tnatt: I think we still want to write the rms to pdm renaming table.. I was more questioning how those mappings would be stored internally, i.e. would we not have rms as the `source_system` similar to stratigraphy mappings?
```
        source_system=DataSystem.rms,
        target_system=DataSystem.pdm,
```

The order of the columns in the renaming table does not matter for the use inside rms 🙂 so if we have `rms` as the first column that works equally fine!

GibranAlfa: 1. if we write rms -> pdm then we will have rms as source system and pdm as target system, i dont get what the problem is 🤔 
2. should we now support rms to pdm and pdm to rms renaming table, or just rms to pdm?

tnatt: I  might be misunderstanding 🤔 but will you be able to find a match in our mappings with the filter you have in the code here? Will we not store a rms to pdm mapping with rms as source_system?

and for number 2 -> nope only rms to pdm

GibranAlfa: because we don't import rms_pdm.csv (because there is no such file as far as i'm aware), we need to map rms to pdm via fmu settings app. If we have that functionality in fmu settings to map rms to pdm (either just change the pdm name from smda mapping, or connect to pdm directly), then the GUI can store mapping from rms to pdm. At least that's how i imagine it, i'm not sure if you are thinking another source system (like storing rms -> smda and then smda -> pdm, so then we can infer rms -> pdm?)

GibranAlfa:
Changes after review:

- Write operations for rms eclipse are changed to rms_simulator.csv and rms_simulator.renaming_table with "SIMULATOR_WELL_NAME" column in csv and "SETNAMES rms\tsimulator" header in renaming table
- Changed write pdm_rms_renaming_table to write rms_pdm_renaming_table
- Write methods now return `None` instead of relative path, and remove the relative path in `_resolve_path()`
- `_write_wellbore_renaming_table()` header is built based on source and target system instead of hardcoded string
- Added test for `_filter_wellbore_mappings()`
- Adjust tests

Not changed:

- still provide default path for write methods

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,266 @@
...
+
+        Raises:
+            ValueError: If the path escapes the project root or no mappings can be
+                represented in the rms_simulator.csv format.
+        """
+        csv_path = self._resolve_path(relative_path, self.RMS_SIMULATOR_CSV_PATH)
+        rows = [
+            {
+                "RMS_WELL_NAME": mapping.source_id,
+                "SIMULATOR_WELL_NAME": mapping.target_id,
```
tnatt: We should standardize this even more to be the target system names `rms` and `simulator`. Maybe we could just do this
```suggestion
                mapping.source_system: mapping.source_id,
                mapping.target_system: mapping.target_id,
```

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,266 @@
...
+if TYPE_CHECKING:
+    from fmu.settings import ProjectFMUDirectory
+
+
+class WellboreMappingsFileIO:
+    """Import and export wellbore mappings from project files outside .fmu."""
+
+    WELL_INFO_DIRECTORY: Final[Path] = Path("rms/input/well_modelling/well_info")
+    RMS_ECLIPSE_CSV_PATH: Final[Path] = WELL_INFO_DIRECTORY / "rms_eclipse.csv"
+    RMS_SIMULATOR_CSV_PATH: Final[Path] = WELL_INFO_DIRECTORY / "rms_simulator.csv"
```
tnatt: I think since we already are creating a new standard, should we include more context in the name here like `rms_simulator_mappings.csv`?

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,266 @@
...
+
+        if not rows:
+            raise ValueError(
+                "No wellbore mappings available to write to rms_simulator.csv"
+            )
+
+        csv_path.parent.mkdir(parents=True, exist_ok=True)
+        with csv_path.open("w", encoding="utf-8", newline="") as file_handle:
+            writer = csv.DictWriter(
+                file_handle, fieldnames=["RMS_WELL_NAME", "SIMULATOR_WELL_NAME"]
```
tnatt: Relates to comment above 🙂 
```suggestion
                file_handle, fieldnames=["rms", "simulator"]
```

src/fmu_settings_api/services/mappings.py
```diff
@@ -41,6 +47,99 @@ def update_internal_stratigraphy_mappings(
...
+            relation_type=InternalRelationType.primary,
+        )
+        if not filtered_wellbore_mappings:
+            raise ValueError(
+                "No rms-to-simulator primary wellbore mappings available to export "
+                "as rms_simulator.renaming_table"
+            )
+        self._wellbore_mappings_file_io.write_rms_simulator_renaming_table(
+            filtered_wellbore_mappings, relative_path
+        )
```
tnatt, line 116: There is still a risk of writing incorrect header on the columns exported in the renaming table, i.e. placing `rms` header on simulator names. We do filtering in one module and do the write in another makes which makes it more difficult to detect.

It could be an option to call the write functionality more directly here. The `write_rms_simulator_renaming_table` does not provide any extra functionality except the resolving of the path,  which could also be done inside this function. Allowing us to remove one wrapping layer. WDYT 🙂 ?

```suggestion
        self._wellbore_mappings_file_io._write_wellbore_renaming_table(
            filtered_wellbore_mappings,
            renaming_table_path=renaming_table_path,
            source_system=DataSystem.rms,
            target_system=DataSystem.pdm,
        )
```

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
GibranAlfa: we can do that

src/fmu_settings_api/services/mappings.py
GibranAlfa: yes, now it's simplified to be one public method in interface that the export methods in mapping service call

GibranAlfa:
also moved the default path to the mapping service. `MappingsService` owns defaults, filtering, which systems to export, and empty-match validation.

`WellboreMappingsFileIO` owns file handling: path resolution, parsing, formatting, and reading/writing the requested file.

src/fmu_settings_api/interfaces/wellbore_mappings_file_io.py
```diff
@@ -0,0 +1,193 @@
...
+        csv_path = self._fmu_dir.resolve_path_inside_project(Path(relative_path))
+        source_system = DataSystem.rms
+        target_system = DataSystem.simulator
+        rows = [
+            {
+                source_system: mapping.source_id,
+                target_system: mapping.target_id,
+            }
+            for mapping in wellbore_mappings
+        ]
```
tnatt, line 138: Still think we should do this to make it robust since the filtering is done outside - there is a risk of adding wrong header to wrong names here.  
```suggestion
        rows = [
            {
                mapping.source_system: mapping.source_id,
                mapping.target_system: mapping.target_id,
            }
            for mapping in wellbore_mappings
        ]
```

tnatt (approved):
Nice work 🙂 just one comment left from me that I think should be addressed.

PR #380, opened 2026-05-28
ENH: Support wellbore mappings in mappings endpoints

### LINKED ISSUE #364: Support wellbore mappings in the mappings endpoints
As we now have support for well mappings in the mappings manager from this issue, https://github.com/equinor/fmu-settings/issues/231, we should support it in the `MappingsService` logics

GibranAlfa:
Resolves #364

- Adds service wrappers for listing and updating internal wellbore mappings.
- Generalizes `MappingsService` read/write-by-source logic across stratigraphy and wellbore mappings.
- Parses PUT mapping payloads from the path mapping type so wellbore and stratigraphy bodies validate against the intended model.

## Checklist


src/fmu_settings_api/v1/routes/project.py
```diff
@@ -1358,19 +1377,26 @@ async def get_mappings(
...
         **LockConflictResponses,
     },
+    openapi_extra=PutMappingsOpenApiExtra,
 )
 async def put_mappings(
     mappings_service: MappingsServiceDep,
     mapping_type: MappingType,
     source_system: DataSystem,
-    mappings: InternalStratigraphyMappings,
+    mappings: Request,
```
slangeveld, line 1386: There is probably a good reason, but can you explain why we cannot set the type of mappings to be an optional `InternalStratigraphyMappings | InternalWellboreMappings` here? Seems like a simpler solution at first glance

GibranAlfa: We could type the body as `InternalStratigraphyMappings | InternalWellboreMappings` but the union cannot be resolved reliably. For example, `[]` is valid for both models, so Pydantic would choose the first match (`InternalStratigraphyMappings`) even for `/mappings/wellbore/rms`.

Since the mapping type is already in the URL, now the route uses it explicitly when validating the body.

The OpenAPI schema also has the same limitation. If we use that union type, the generated TypeScript type will allow both type, even though if you call `PUT /mappings/wellbore/rms`, only  `InternalWellboreMappings` type should be allowed. That's why we should check the request body type based on the path parameter (in `_parse_internal_mappings_payload`).

Also `InternalStratigraphyMappings | InternalWellboreMappings` is not optional, it required but can be one of them, optional case would be `InternalStratigraphyMappings | InternalWellboreMappings | None`.

src/fmu_settings_api/services/mappings.py
```diff
@@ -169,53 +187,92 @@ def get_internal_mappings_by_source_system(
...
+            )
+
+        if any(mapping.source_system != source_system for mapping in mappings):
+            raise ValueError(
+                "All mappings in the request body must use the requested "
+                f"source system '{source_system.value}'"
+            )
+
+        try:
+            existing_mappings = list_mappings()
```
slangeveld: Can we call this `existing_mappings_for_mapping_type` or similar to clearify that this does not actually return all mappings?

src/fmu_settings_api/services/mappings.py
```diff
@@ -169,53 +187,92 @@ def get_internal_mappings_by_source_system(
...
+        except FileNotFoundError:
+            existing_mappings = mappings_model.model_validate([])
+
+        other_mappings = [
```
slangeveld: Here as well, it would be good to clarify in the variable name that this is only a subset of the mappings from the mappings file

src/fmu_settings_api/services/mappings.py
```diff
@@ -169,53 +187,92 @@ def get_internal_mappings_by_source_system(
...
-                    for mapping in stratigraphy_mappings
-                    if mapping.source_system == source_system
-                ]
-            )
-            return InternalMappings(stratigraphy=filtered_mappings)
+        mappings_model, list_mappings, _ = (
+            self._resolve_internal_mapping_implementation(mapping_type)
+        )
+        try:
+            mappings = list_mappings()
```
slangeveld: Maybe we should call this mappings_for_mapping_type or similar to clearify that this does not actually return all mappings?

src/fmu_settings_api/v1/routes/project.py
slangeveld: Thanks for the explanation 👍

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -1460,3 +1486,25 @@ def _create_opened_project_response(fmu_dir: ProjectFMUDirectory) -> FMUProject:
         raise HTTPException(
             status_code=403, detail="Permission denied accessing .fmu"
         ) from e
+
+
+def _parse_internal_mappings_payload(
+    mapping_type: MappingType,
+    payload: object,
+) -> InternalStratigraphyMappings | InternalWellboreMappings:
+    """Parse the request body as the mappings type from the URL.
```
slangeveld: At this point the `mapping_type` has already been extracted from the URL so I think we should maybe avoid mentioning URL here and only refer to the `mapping_type`. 

I think the explanation here of how the URL says whether this is stratigraphy or wellbore is good, but it belongs in the docstring of the `put_mappings`method above where this extraction actually is done.

tests/test_services/test_mappings_service.py
```diff
@@ -281,6 +432,65 @@ def test_update_internal_mappings_by_source_system_rejects_mismatched_source_sys
...
+            InternalStratigraphyMappings,
+            "list_internal_stratigraphy_mappings",
+            "update_internal_stratigraphy_mappings",
+        ),
+        (
+            MappingType.wellbore,
+            InternalWellboreMappings,
+            "list_internal_wellbore_mappings",
+            "update_internal_wellbore_mappings",
+        ),
```
slangeveld, line 455: Maybe also include a case here where the mapping_type is of an unsupportable type?

tests/test_v1/test_project.py
```diff
@@ -3241,6 +3242,32 @@ async def test_put_mappings_stratigraphy_success(
     )
```
slangeveld, line 3244: Maybe we should also include a test here where the `mapping_type` in the URL does not match the mapping object it will try to resolve it to? For example if a wellbore payload is provided in the request while the mapping_type in the url path is `stratigraphy`

slangeveld (approved):
Nice 👍  A few comments regarding variable names, docstrings and tests to consider

GibranAlfa:
thanks for the feedbacks @slangeveld, i agree with all and addressed them 👍

PR #383, opened 2026-06-01
ENH: Use Drogon masterdata in SMDA endpoint for Drogon

### LINKED ISSUE #381: Serve dummy masterdata/stratigraphy for Drogon
The Drogon model can't use FMU settings since Drogon is not a field in SMDA. 
Since Drogon is an training model used in courses etc. it is important that FMU settings works for Drogon.

For relevant SMDA endpoints, instead of connecting to the database we should serve Drogon mock data for masterdata and stratigraphy when Drogon is the field.

We have stored relevant Drogon data here:
https://github.com/equinor/fmu-settings/blob/77e81491ea27424af4a3bdc410501ae9a6d0ccea/src/fmu/settings/_drogon/_data.py#L13

GibranAlfa:
Resolves #381 

- Serve Drogon field search and masterdata directly from `fmu.settings._drogon` instead of querying SMDA.
- Support Drogon masterdata lookup by field name or UUID.
- Serve Drogon stratigraphic units locally from the Drogon mappings.

image
image
image
 There's an issue that we can't map TopVolon to Volon Fm. Top because it has the same uuid as Therys Fm. Top in Drogon data. Created this issue to fix that equinor/fmu-settings/issues/261

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/smda.py
+++ b/src/fmu_settings_api/services/smda.py
@@ -72,7 +72,13 @@ DROGON_STRATIGRAPHIC_UNITS: Final[list[StratigraphicUnit]] = [
 
 def _is_drogon_identifier(identifier: str) -> bool:
     """Returns whether an identifier refers to the Drogon field."""
-    return identifier.casefold() == DROGON_FIELD["identifier"].casefold()
+    search_identifier = identifier.casefold()
+    drogon_identifier = DROGON_FIELD["identifier"].casefold()
+
+    if search_identifier.endswith("*"):
+        return drogon_identifier.startswith(search_identifier.removesuffix("*"))
+
+    return search_identifier == drogon_identifier
 
 
 def _is_drogon_masterdata_request(smda_fields: list[SmdaSelectedField]) -> bool:
--- a/src/fmu_settings_api/services/smda.py
+++ b/src/fmu_settings_api/services/smda.py
@@ -2,6 +2,7 @@
 
 import asyncio
 from collections.abc import Sequence
+from typing import Any, Final
 
 import httpx
 from fmu.datamodels.common.masterdata import (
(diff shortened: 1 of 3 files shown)


tnatt (approved):
Really nice work :) Look forward to demoing this!

Btw, would we like to support wildcards for Drogon, like finding Drogon if the users type `DRO*`? I guess that does not work now, and I am not sure we need to add that complexity either.

I think we should also include a mapping for the `TopVolantis_BaseVolantis` zone from rms, That should be mapped to `Volantis Gp.`. But this needs to be added to `fmu-settings,` and is unrelated to this PR.

GibranAlfa:
we now support Drogon wildcard
image
yes, when that change is made in `fmu-settings` we can map it in FMU settings app.

tnatt:
@GibranAlfa with adding that wildcard support, we exclude other SMDA fields that starts with the same letters.
E.g. typing `DRO*` will e.g. exclude the following fields 

- DROBYSHEVSKOYE
- DROBYSHIVSKE
- DROSHKY
- DROZDOVSKOYE VOSTOCHNOYE

GibranAlfa:
> @GibranAlfa with adding that wildcard support, we exclude other SMDA fields that starts with the same letters.
> 
> E.g. typing `DRO*` will e.g. exclude the following fields 
> 
> 
> 
> - DROBYSHEVSKOYE
> 
> - DROBYSHIVSKE
> 
> - DROSHKY
> 
> - DROZDOVSKOYE VOSTOCHNOYE

Yes I'm aware, are those fields relevant though? I tried one of them and they don't have coordinate systems so it can't be used anyway

tnatt:
But what about D* - then we exclude all SMDA field starting with the character D, and return only DROGON

GibranAlfa:
Hmm that's true, I guess we don't need to support wildcard then

tnatt:
I agree, for now we don't and then we can add it later if necessary. Otherwise you would have to serve a mix of Drogon results with actual SMDA results. Which is not too difficult, but lets add an issue for it and do it in a separate PR.

GibranAlfa:
yes we can mix their results, I'll create a separate issue for that

PR #389, opened 2026-06-04
ENH: Simplify match endpoint to generic ranked name matching

### LINKED ISSUE #378: Improvement to the matching service
Some questions

1. Should we change the matching endpoint to accept things to match (zones, horizons, wells from RMS project) and things to be matched (SMDA strat units, wellbore identifier, etc) so the endpoint is general and the GUI decide what it wants to match? Right now the endpoint automatically take zones in project config, search for strat units in SMDA, and do the matching in one endpoint. I prefer we make it general.
2. Should the matching be strict (Volantis Gp will not be 100% match with Gp Volantis) or not strict (Volantis Gp == Gp Volantis)? I guess we can make it strict? Not sure if there's any case that ppl have their RMS zone/horizon names in opposite order from SMDA name
3. Should we let users to add rules before matching? Currently we normalize the names before matching (converts to lowercase, replaces underscores and dots with spaces. Maybe we should remove dash and slash as well?).

Image

I think we can:
   - Let user input strings to remove (?)
   - Let user input strings to replace, for ex. replace GP with Group
   - other things?

GibranAlfa:
Resolves #378

Simplified the match endpoint into a generic name matching utility. The endpoint now accepts caller provided sources, targets, and optional replacement rules, returns grouped top 3 candidate matches per source, and no longer reads project config or calls SMDA.

Removed coordinate-system matching from the match API and replaced RMS/SMDA-specific match models with generic request/result models.

## Example

Request:

```json
{
  "sources": ["Viking GP", "Tarbert_Fm"],
  "targets": [
    "Viking Group",
    "Viking Formation",
    "Tarbert Formation",
    "Unrelated Unit"
  ],
  "replacements": [
    { "original": "GP", "replacement": "Group" },
    { "original": "Fm", "replacement": "Formation" }
  ]
}
```

Result:

```json
[
  {
    "source": "Viking GP",
    "matches": [
      {
        "target": "Viking Group",
        "score": 100,
        "confidence": "high"
      },
      {
        "target": "Viking Formation",
        "score": 64.28571428571428,
        "confidence": "medium"
      },
      {
        "target": "Unrelated Unit",
        "score": 23.076923076923073,
        "confidence": "low"
      }
    ]
  },
  {
    "source": "Tarbert_Fm",
    "matches": [
      {
        "target": "Tarbert Formation",
        "score": 100,
        "confidence": "high"
      },
      {
        "target": "Viking Formation",
        "score": 60.60606060606061,
        "confidence": "medium"
      },
      {
        "target": "Unrelated Unit",
        "score": 32.25806451612904,
        "confidence": "low"
      }
    ]
  }
]
```

## Checklist

### DIFF
--- a/src/fmu_settings_api/v1/routes/match.py
+++ b/src/fmu_settings_api/v1/routes/match.py
@@ -3,74 +3,19 @@
 from textwrap import dedent
 from typing import Final
 
-import httpx
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter
 
-from fmu_settings_api.config import HttpHeader
-from fmu_settings_api.deps import ProjectSmdaServiceDep
 from fmu_settings_api.deps.match import MatchServiceDep
-from fmu_settings_api.deps.session import ProjectSessionDep
-from fmu_settings_api.models.match import (
-    RmsCoordinateSystemMatch,
-    RmsStratigraphyMatch,
-)
-from fmu_settings_api.v1.responses import (
-    GetSessionResponses,
-    Responses,
-    inline_add_response,
-)
+from fmu_settings_api.models.match import MatchRequest, MatchResult
+from fmu_settings_api.v1.responses import Responses, inline_add_response
 
 MatchResponses: Final[Responses] = {
     **inline_add_response(
-        400,
-        dedent(
-            """
-            Required configuration is missing from the project config,
-            or invalid parameters are provided.
-            """
-        ),
-        [
-            {"detail": "RMS zones not found in project configuration"},
-            {"detail": "RMS coordinate system not found in project configuration"},
-            {"detail": "SMDA masterdata not found in project configuration"},
-            {"detail": "Stratigraphic column identifier not found in masterdata"},
-        ],
-    ),
-    **inline_add_response(
         422,
-        dedent(
-            """
-            SMDA returns valid data but no results are found,
-            or configuration exists but contains no matchable data.
-            """
-        ),
+        "The request body is missing required match input or contains invalid data.",
         [
-            {"detail": "No stratigraphic units found for column: {identifier}"},
-            {"detail": "No RMS zones available for matching"},
-        ],
-    ),
-    **inline_add_response(
-        500,
-        dedent(
-            """
-            SMDA returns a malformed response that doesn't match
-            the expected structure.
-            """
-        ),
-        [
-            {"detail": "Malformed response from SMDA: {error_details}"},
-        ],
-    ),
-    **inline_add_response(
-        503,
-        dedent(
-            """
-            An API call to SMDA times out or the service is unavailable.
-            """
-        ),
-        [
-            {"detail": "SMDA API request timed out. Please try again."},
-            {"detail": "SMDA error requesting {url}"},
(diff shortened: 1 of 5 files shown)


src/fmu_settings_api/v1/routes/match.py
```diff
@@ -3,207 +3,63 @@
...
+        The endpoint is a pure matching utility. Callers provide both the
+        source names and target names, and can optionally provide string
+        replacement rules to apply before matching.
 
-        This endpoint automatically fetches:
-        - RMS coordinate system from the project configuration
-        - SMDA coordinate system from the project's masterdata configuration
+        Names are normalized before matching by lowercasing, replacing
+        underscores, dots, dashes, and slashes with spaces, collapsing
+        whitespace, and applying optional replacement rules. Replacement
```
GibranAlfa, line 40: we should tell this in the GUI that the names are always normalized first by lowercasing etc, so users don't need to input - or / to replace with "", if they do that it will fail validation with error 
msg:
``` 
"Original must include text after normalization. Empty values and separators such as "_", ".", "-", and "/" cannot be replaced. Those separators are normalized automatically. Replacement rules 
are only needed for text values, for example "Fm" -> "Formation".
```
This also to prevent user inputing empty string in text to replace and put something for replacement, e.g. "" -> "group".

joargr:
Good idea to have a general name matching function. But I wonder if there should be more logic used in the matching endpoint. For instance, there would probably be a need for different logic when matching zones, horizons and wellbores.

When matching a horizon, that is, finding the corresponding SMDA horizon for an RMS horizon, if the user has mapped the zones then the SMDA zones would indicate the top and base horizon for each zone. Using this information an SMDA horizon could be matched. For horizons with multiple zones the logic would have to find the best zone (most coarse zone?) to consult. When the backend itself is checking the project data as well as data from the SMDA API the backend has the information needed for the logic.

When matching wellbores I imagine that an important rule would be to ignore country prefixes of wellbore names. It would be good if the client wouldn't have to supply a list of all relevant country prefixes (sent as `replacements` data), but let this be handled by the backend. Then the client would only have to tell if the rule (named, say, `ignore-country-prefix`) should apply or not, and the backend would know which strings to ignore.

I think it would be good to not rely only on a general name matching function, but to specify the category or type of names (zones, horizons or wellbores) that needs to be matched, so that the backend can use available information to improve the matching. Then the client sends a list of the names to match, along with which rules to apply, but without additional information for the backend to use for the lookup. But this is something that can be a topic for discussion in a meeting.

GibranAlfa:
I agree with the suggestion. I have thought about having different logic for different mapping types as well. But I think this PR can keep the generic name-matching endpoint as a small utility, and we can add domain-aware matching in a follow-up PR once we settle the semantics for stratigraphy/wellbores.

The horizon case especially may need project mappings and SMDA data, so it needs a separate endpoint that requires project session (in this generic endpoint, no session required because it's just a utility function). We can still let the GUI use the generic endpoint now, then introduce the richer matching flow when the backend logic is defined.

joargr (approved):
This looks good, and seems to work nicely as a name matching service.

PR #393, opened 2026-06-08
ENH: Handle missing RMS project errors in POST /rms

### LINKED ISSUE #392: `POST /rms/` does not consistently map RMS open failures to useful HTTP errors
`POST /api/v1/rms/` has several error-handling gaps when opening an RMS project fails. Some expected RMS/project errors can leak as generic `500` responses or produce misleading secondary errors instead of returning a clear API response.

**Problems**

1. `FileNotFoundError` is not caught

`RmsService.get_rms_version()` normalizes missing RMS project paths and missing `.master` files into `FileNotFoundError`, but `post_rms_project()` route does not catch it.

Examples:
- RMS project path does not exist
- RMS project `.master` file does not exist

Expected:
- Return `404` with a clear message.

Actual:
- Falls through to generic error handling.

2. `RmsVersionError` can trigger an unbound local error

When no RMS version is supplied in the request, the route assigns `version` from:

```python
version = rms_service.get_rms_version(rms_project_path)
```

If `get_rms_version()` raises `RmsVersionError`, then `version` is never assigned. The `except RmsVersionError` block then references `version`, which can cause an `UnboundLocalError`.

Expected:
- Return the existing `422` RMS version error response.

Actual:
- Can produce a secondary `500` due to `version` being undefined.

**Suggested Fix**

Update `post_rms_project()` to:

- initialize `version` before the `try` block
- catch `FileNotFoundError` and return `404`

GibranAlfa:
Resolves #392

This updates `POST /rms/` error handling so missing RMS project paths, missing `.master` files, and unsupported detected RMS versions are mapped to useful HTTP responses.

The route now avoids starting the RMS API worker when an explicit RMS version is supplied but the configured RMS project path does not exist. It also preserves 404 handling for missing project metadata while detecting the version, and documents the additional 404 responses.

With this change, we will see this:
image
There's a duplicate error for now because the GUI retries when it can't open RMS project. Made this issue to fix that https://github.com/equinor/fmu-settings-gui/issues/430

## Checklist

- [x] Checked the boxes in this checklist

### DIFF
--- a/src/fmu_settings_api/services/rms.py
+++ b/src/fmu_settings_api/services/rms.py
@@ -39,7 +39,7 @@ class RmsService:
             rms_config = RmsConfig(project=str(rms_project_path))
         except RmsProjectNotFoundError as e:
             raise FileNotFoundError(
-                f"RMS project not found at '{rms_project_path}': {str(e)}"
+                f"RMS project not found at '{rms_project_path}'."
             ) from e
         except FileNotFoundError as e:
             raise FileNotFoundError(
--- a/src/fmu_settings_api/v1/routes/rms.py
+++ b/src/fmu_settings_api/v1/routes/rms.py
@@ -61,12 +61,18 @@ FailedOpeningRmsProjectResponses: Final[Responses] = {
     ),
     **inline_add_response(
         404,
-        dedent("RMS project does not exist at the configured path."),
+        dedent("RMS project or its .master file was not found."),
         [
             {
                 "detail": (
                     "RMS project does not exist at the "
                     "configured path: {rms_project_path}."
+                )
+            },
+            {
+                "detail": (
+                    "RMS project .master file not found at "
+                    "'{rms_project_path}': {string content of exception}"
                 )
             },
         ],
@@ -121,12 +127,15 @@ async def post_rms_project(
     rms_version: RmsVersion | None = None,
 ) -> Message:
     """Open an RMS project and store it in the session."""
-    try:
-        version = (
-            rms_version.version
-            if rms_version is not None
-            else rms_service.get_rms_version(rms_project_path)
+    version = rms_version.version if rms_version is not None else None
+    if version is not None and not rms_project_path.exists():
+        raise HTTPException(
+            status_code=404,
+            detail=f"RMS project not found at '{rms_project_path}'.",
         )
+    try:
+        if version is None:
+            version = rms_service.get_rms_version(rms_project_path)
         executor, project = rms_service.open_rms_project(rms_project_path, version)
         await session_service.add_rms_session(executor, project)
         return Message(
@@ -142,6 +151,8 @@ async def post_rms_project(
                 f"configured path: {rms_project_path}."
             ),
         ) from e
+    except FileNotFoundError as e:
+        raise HTTPException(status_code=404, detail=str(e)) from e
     except RmsVersionError as e:
         raise HTTPException(
             status_code=422,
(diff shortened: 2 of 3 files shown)


src/fmu_settings_api/services/rms.py
```diff
@@ -39,7 +39,7 @@ def get_rms_version(rms_project_path: Path) -> str:
             rms_config = RmsConfig(project=str(rms_project_path))
         except RmsProjectNotFoundError as e:
             raise FileNotFoundError(
-                f"RMS project not found at '{rms_project_path}': {str(e)}"
+                f"RMS project not found at '{rms_project_path}'."
```
GibranAlfa, line 42: remove the error msg because it will say something like
```
Error accessing the RMS project: RMS project not found at '/path/to/project.rms13.1.0': The project: /path/to/project.rms13.1.0 does not exist as a directory.
```
That does not exist as a dir is a redundant error msg.

src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -121,12 +128,15 @@ async def post_rms_project(
...
 ) -> Message:
     """Open an RMS project and store it in the session."""
-    try:
-        version = (
-            rms_version.version
-            if rms_version is not None
-            else rms_service.get_rms_version(rms_project_path)
+    version = rms_version.version if rms_version is not None else None
+    if version is not None and not rms_project_path.exists():
+        raise HTTPException(
```
GibranAlfa, line 132: we stop early when rms version is given but rms project does not exists, so we don't try to start the RMS proxy

src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -121,12 +127,15 @@ async def post_rms_project(
...
+            status_code=404,
+            detail=f"RMS project not found at '{rms_project_path}'.",
         )
+    try:
+        if version is None:
+            version = rms_service.get_rms_version(rms_project_path)
```
GibranAlfa, line 138: Here if this project path doesn't exist it will raise `FileNotFoundError`, will be caught in except block below.

src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -61,14 +61,20 @@
...
         [
             {
                 "detail": (
                     "RMS project does not exist at the "
                     "configured path: {rms_project_path}."
                 )
             },
+            {
+                "detail": (
+                    "RMS project .master file not found at "
```
slangeveld, line 74: Probably fine, just asking the question: Is this information about the .master file to technical for the user to understand?

src/fmu_settings_api/v1/routes/rms.py
```diff
@@ -142,6 +151,8 @@ async def post_rms_project(
                 f"configured path: {rms_project_path}."
             ),
         ) from e
+    except FileNotFoundError as e:
+        raise HTTPException(status_code=404, detail=str(e)) from e
```
slangeveld, line 155: There should be a test for this case where the project path doesn't exist, version is None and it tries to get the version

slangeveld (approved):
Code looks good 👍  Some minor comments. 

@krhart should probably also make a review to add here thoughts from a user perspective

src/fmu_settings_api/v1/routes/rms.py
GibranAlfa: probably ok, this error is coming from `runrms`

PR #397, opened 2026-06-12
ENH: Improve RMS .master missing error message

### LINKED ISSUE #396: Use better error message if RMS .master file is not found
We use RMS .master file to automatically determine what RMS version is the project running. The users currently will see error message if .master file is not found, and thus we can't determine the RMS version automatically. All users might not be familiar with .master file, so we can tell that RMS version cannot be determined in the error message in addition to master file is not found, because it's still a useful thing to know.

GibranAlfa:
Resolves #396 

Improved RMS .master missing error message by adding the RMS version cannot be determined when .master file is missing.

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/rms.py
+++ b/src/fmu_settings_api/services/rms.py
@@ -43,7 +43,8 @@ class RmsService:
             ) from e
         except FileNotFoundError as e:
             raise FileNotFoundError(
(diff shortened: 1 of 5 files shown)


krhart (approved):
A lot more "human readable" error messages 👌

PR #403, opened 2026-06-19
ENH: Add filter for changelog

### LINKED ISSUE #296: Add filtertype to the changelog endpoint
Add filtertype to changelog endpoint.

nourinmohd:
Resolves #296 

Added filtertype to changelog endpoint. 

## Checklist

### DIFF
--- a/src/fmu_settings_api/deps/changelog.py
+++ b/src/fmu_settings_api/deps/changelog.py
@@ -5,6 +5,7 @@ from typing import Annotated, Literal
 
 from fastapi import Depends, HTTPException, Query
 from fmu.settings.models._enums import ChangeType, FilterType
+from fmu.settings.models.change_info import ChangeInfo
 from fmu.settings.models.log import Filter
 
 from fmu_settings_api.deps.session import ProjectSessionDep
@@ -48,6 +49,11 @@ async def get_changelog_filters(
         assert filter_value is not None
         assert filter_type is not None
         assert operator is not None
+        if field_name not in ChangeInfo.model_fields:
+            raise HTTPException(
+                status_code=422,
+                detail=f"Unknown changelog field: {field_name}.",
+            )
         filter_ = Filter(
             field_name=field_name,
             filter_value=filter_value,
--- a/src/fmu_settings_api/deps/changelog.py
+++ b/src/fmu_settings_api/deps/changelog.py
@@ -11,7 +11,7 @@ from fmu_settings_api.deps.session import ProjectSessionDep
 from fmu_settings_api.services.changelog import ChangelogService
 
 
-@dataclass
+@dataclass(frozen=True)
 class ChangelogFilters:
     """Query filters for changelog retrieval."""
 
@@ -22,7 +22,7 @@ class ChangelogFilters:
 
 async def get_changelog_filters(
     change_type: ChangeType | None = None,
-    max_entries: Annotated[int | None, Query(ge=1)] = None,
+    max_entries: int | None = Query(default=None, ge=1),
     field_name: str | None = None,
     filter_value: str | None = None,
     filter_type: FilterType | None = None,
--- a/src/fmu_settings_api/deps/changelog.py
+++ b/src/fmu_settings_api/deps/changelog.py
@@ -5,7 +5,6 @@ from typing import Annotated, Literal
 
 from fastapi import Depends, HTTPException, Query
 from fmu.settings.models._enums import ChangeType, FilterType
-from fmu.settings.models.change_info import ChangeInfo
 from fmu.settings.models.log import Filter
 
 from fmu_settings_api.deps.session import ProjectSessionDep
@@ -30,16 +29,20 @@ async def get_changelog_filters(
     operator: Literal[">", ">=", "<", "<=", "==", "!="] | None = None,
 ) -> ChangelogFilters:
     """Return parsed changelog query filters."""
-    generic_filter_values = (field_name, filter_value, filter_type, operator)
-    has_any_generic_filter = any(value is not None for value in generic_filter_values)
-    has_all_generic_filter = all(value is not None for value in generic_filter_values)
+    generic_log_filter_values = (field_name, filter_value, filter_type, operator)
+    has_any_generic_filter = any(
+        value is not None for value in generic_log_filter_values
+    )
+    has_all_generic_filter = all(
+        value is not None for value in generic_log_filter_values
+    )
 
     if has_any_generic_filter and not has_all_generic_filter:
         raise HTTPException(
             status_code=422,
             detail=(
-                "Generic changelog filter requires all of: field_name, "
-                "filter_value, filter_type, operator."
+                "Changelog filtering requires all of the generic log filtering "
+                "fields: field_name, filter_value, filter_type, operator."
             ),
         )
 
@@ -49,11 +52,6 @@ async def get_changelog_filters(
         assert filter_value is not None
         assert filter_type is not None
         assert operator is not None
-        if field_name not in ChangeInfo.model_fields:
-            raise HTTPException(
-                status_code=422,
-                detail=f"Unknown changelog field: {field_name}.",
-            )
         filter_ = Filter(
             field_name=field_name,
             filter_value=filter_value,
--- a/src/fmu_settings_api/deps/changelog.py
+++ b/src/fmu_settings_api/deps/changelog.py
@@ -1,11 +1,65 @@
 """Changelog service dependencies."""
 
-from typing import Annotated
+from dataclasses import dataclass
+from typing import Annotated, Literal
 
-from fastapi import Depends
+from fastapi import Depends, HTTPException, Query
+from fmu.settings.models._enums import ChangeType, FilterType
+from fmu.settings.models.log import Filter
 
 from fmu_settings_api.deps.session import ProjectSessionDep
 from fmu_settings_api.services.changelog import ChangelogService
+
+
+@dataclass
+class ChangelogFilters:
+    """Query filters for changelog retrieval."""
+
+    change_type: ChangeType | None = None
+    max_entries: int | None = None
+    filter_: Filter | None = None
+
+
+async def get_changelog_filters(
+    change_type: ChangeType | None = None,
+    max_entries: Annotated[int | None, Query(ge=1)] = None,
+    field_name: str | None = None,
+    filter_value: str | None = None,
+    filter_type: FilterType | None = None,
+    operator: Literal[">", ">=", "<", "<=", "==", "!="] | None = None,
+) -> ChangelogFilters:
+    """Return parsed changelog query filters."""
+    generic_filter_values = (field_name, filter_value, filter_type, operator)
+    has_any_generic_filter = any(value is not None for value in generic_filter_values)
+    has_all_generic_filter = all(value is not None for value in generic_filter_values)
+
+    if has_any_generic_filter and not has_all_generic_filter:
+        raise HTTPException(
+            status_code=422,
+            detail=(
+                "Generic changelog filter requires all of: field_name, "
+                "filter_value, filter_type, operator."
+            ),
+        )
+
+    filter_: Filter | None = None
+    if has_all_generic_filter:
+        assert field_name is not None
+        assert filter_value is not None
+        assert filter_type is not None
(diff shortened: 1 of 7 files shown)


GibranAlfa (commented):
I think we should expand on this. We should be able to filter based on:

1. `ChangeType`, like you did in this PR
2. Generic log filtering using the existing `fmu.settings.models.log.Filter` [model](https://github.com/equinor/fmu-settings/blob/51c887c185db686de309c60adf7be3e6970381dc/src/fmu/settings/models/log.py#L36-L56), with `field_name`, `filter_value`, `filter_type`, and `operator`. Since `Filter` requires all four fields, the API query params can be optional individually, but should return 422 if only some of them are provided.
3. A separate `max_entries`/entry-count query param for the amount of returned entries, so in GUI homepage we can just return 5 latest entries instead of loading the whole changelog and filter it in GUI.

To keep the endpoint from getting too many query params directly in the signature, maybe add a changelog query dependency in `deps/changelog.py`, for example `ChangelogFiltersDep`. That dependency can collect `change_type`, `max_entries`, and the optional generic `Filter` params, build the upstream `Filter` only when all generic filter params are provided, and raise 422 when the generic filter is partial.

The `ChangelogService.get_changelog()` should also use built-in [method](https://github.com/equinor/fmu-settings/blob/51c887c185db686de309c60adf7be3e6970381dc/src/fmu/settings/_resources/log_manager.py#L47) `self._fmu_dir.changelog.filter_log(filter_)` to serve the filtering.

No need to have `_filter_by_change_type()` helper, just inline them in `ChangelogService.get_changelog()`.

src/fmu_settings_api/deps/changelog.py
```diff
@@ -1,13 +1,67 @@
...
-from fastapi import Depends
+from fastapi import Depends, HTTPException, Query
+from fmu.settings.models._enums import ChangeType, FilterType
+from fmu.settings.models.log import Filter
 
 from fmu_settings_api.deps.session import ProjectSessionDep
 from fmu_settings_api.services.changelog import ChangelogService
 
 
+@dataclass
```
GibranAlfa: better to make this immutable as it's just parsed from request data 
```suggestion
@dataclass(frozen=True)
```

src/fmu_settings_api/deps/changelog.py
```diff
@@ -1,13 +1,67 @@
...
+    """Query filters for changelog retrieval."""
+
+    change_type: ChangeType | None = None
+    max_entries: int | None = None
+    filter_: Filter | None = None
+
+
+async def get_changelog_filters(
+    change_type: ChangeType | None = None,
+    max_entries: Annotated[int | None, Query(ge=1)] = None,
```
GibranAlfa: this can be simplified
```suggestion
    max_entries: int | None = Query(default=None, ge=1),
```

tests/test_v1/test_project.py
```diff
@@ -335,6 +335,152 @@ async def test_get_changelog_success(
     assert response_data[0]["key"] == "changelog_test"
 
 
+async def test_get_changelog_success_with_filtertype(
```
GibranAlfa: this should be renamed now it filter based on change type and not filter type

tests/test_v1/test_project.py
```diff
@@ -335,6 +335,152 @@ async def test_get_changelog_success(
...
+        )
+
+    response = client_with_project_session.get(f"{ROUTE}/changelog?max_entries=2")
+
+    assert response.status_code == status.HTTP_200_OK
+    response_data = response.json()
+    assert len(response_data) == 2
+    assert [entry["key"] for entry in response_data] == ["entry_2", "entry_3"]
+
+
```
GibranAlfa, line 336: also better to add test when all the filtering are used (generic filter, change type, max_entries), and case where max_entries=0 should raise 422

src/fmu_settings_api/v1/routes/project.py
GibranAlfa, line 386: now that we have a new 422 case (invalid query parameters), this should be updated

src/fmu_settings_api/deps/changelog.py
```diff
@@ -1,13 +1,67 @@
...
+
+    change_type: ChangeType | None = None
+    max_entries: int | None = None
+    filter_: Filter | None = None
+
+
+async def get_changelog_filters(
+    change_type: ChangeType | None = None,
+    max_entries: int | None = Query(default=None, ge=1),
+    field_name: str | None = None,
```
GibranAlfa, line 26: we need to validate `field_name` to ensure it provides the correct column name so Pandas can access it correctly. Add this before constructing `Filter`
```py
if field_name not in ChangeInfo.model_fields:
    raise HTTPException(
        status_code=422,
        detail=f"Unknown changelog field: {field_name}.",
    )
```

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -384,10 +384,27 @@
...
+                "detail": [
+                    {
+                        "type": "greater_than_equal",
+                        "loc": ["query", "max_entries"],
+                        "msg": "Input should be greater than or equal to 1",
+                        "input": "0",
+                        "ctx": {"ge": 1},
+                    }
+                ]
+            },
```
GibranAlfa: we dont need to document generic pydantic error case like this, because FastAPI already documents the standard 422 validation response, we can remove this

tests/test_v1/test_project.py
```diff
@@ -335,6 +335,211 @@ async def test_get_changelog_success(
...
+    assert session_id is not None
+    session = await get_fmu_session(session_id)
+    assert isinstance(session, ProjectSession)
+
+    fmu_dir = session.project_fmu_directory
+    for change_type, key in (
+        (ChangeType.update, "entry_1"),
+        (ChangeType.update, "entry_2"),
+        (ChangeType.remove, "entry_3"),
+        (ChangeType.update, "entry_4"),
```
GibranAlfa: because this test filter by the user, add an entry for other user as well, so we know that filtering by user wont include the other user,
```py
for change_type, user, key in (
    (ChangeType.update, "test_user", "entry_1"),
    (ChangeType.update, "test_user", "entry_2"),
    (ChangeType.remove, "test_user", "entry_3"),
    (ChangeType.update, "other_user", "excluded_entry"),
    (ChangeType.update, "test_user", "entry_4"),
):
    ...
```

src/fmu_settings_api/v1/responses.py
```diff
@@ -8,7 +8,7 @@
 
 
 def add_response_example(
-    responses: Responses, http_code: int, example: dict[str, str]
+    responses: Responses, http_code: int, example: dict[str, Any]
```
GibranAlfa: change this back if you remove the pydantic error doc

GibranAlfa (commented):
why are you asking for review again? @nourinmohd

src/fmu_settings_api/deps/changelog.py
```diff
@@ -1,13 +1,67 @@
...
+    filter_value: str | None = None,
+    filter_type: FilterType | None = None,
+    operator: Literal[">", ">=", "<", "<=", "==", "!="] | None = None,
+) -> ChangelogFilters:
+    """Return parsed changelog query filters."""
+    generic_filter_values = (field_name, filter_value, filter_type, operator)
```
slangeveld: Maybe we should name this `generic_log_filter_values` to make it clear in which context they are generic?

src/fmu_settings_api/deps/changelog.py
```diff
@@ -1,13 +1,73 @@
...
+    has_any_generic_filter = any(value is not None for value in generic_filter_values)
+    has_all_generic_filter = all(value is not None for value in generic_filter_values)
+
+    if has_any_generic_filter and not has_all_generic_filter:
+        raise HTTPException(
+            status_code=422,
+            detail=(
```
slangeveld, line 43: We can probably make this error message a bit clearer to the user as "Generic changelog filter" might not make so much sense to them. Maybe something like 
`Changelog filtering requires all of the generic log filtering fields: field_name, filter_value, filter_type, operator.` or similar

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -384,10 +384,17 @@
...
         422,
-        "Invalid changelog data",
+        "Invalid changelog data or query parameters",
         [
             {"detail": "Invalid changelog format or data at {path}: {error}"},
             {"detail": "Invalid or corrupt JSON at {path}: {error}"},
+            {"detail": "Unknown changelog field: {field_name}."},
+            {
+                "detail": (
+                    "Generic changelog filter requires all of: field_name, "
```
slangeveld: If we change the error message we should also change it here

tests/test_services/test_changelog_service.py
```diff
@@ -32,3 +33,103 @@ def test_get_changelog_contains_changeinfo_entries(
...
+                hostname="localhost",
+                file="config.json",
+                key=key,
+            )
+        )
+
+    result = service.get_changelog(max_entries=2)
+
+    assert len(result) == 2
+    assert [entry.key for entry in result] == ["entry_2", "entry_3"]
```
slangeveld, line 134: There should probably also be a test here where all of the filter options `change_type`, the generic log filter and `max_entries` are filtered on at the same time, and that the expected subset og log entries are returned.

I see you have that in the test_project.py file. You can move that to here.

tests/test_v1/test_project.py
```diff
@@ -335,6 +335,228 @@ async def test_get_changelog_success(
     assert response_data[0]["key"] == "changelog_test"
 
 
+async def test_get_changelog_success_with_change_type(
```
slangeveld: All the tests here are already covered by the tests in `test_changelog_services`. 

The test here of the GET changelog route should only test
1. The success case: That `changelog_service.get_changelog()` is called and returns _something_. What it returns (the internal logic of the get_changelog()) is none of the route layers concern and should not be tested here.
2. The exception cases: That the expected HTTPException is raised for the different exceptions

src/fmu_settings_api/deps/changelog.py
```diff
@@ -16,3 +76,4 @@ async def get_changelog_service(
 
 
 ChangelogServiceDep = Annotated[ChangelogService, Depends(get_changelog_service)]
+ChangelogFiltersDep = Annotated[ChangelogFilters, Depends(get_changelog_filters)]
```
slangeveld, line 77: This new dependency `ChangelogFiltersDep` and its method `get_changelog_filters()` does not have any tests at the moment. That should be added to the test_deps.py file.

src/fmu_settings_api/deps/changelog.py
```diff
@@ -1,13 +1,73 @@
...
+    if has_all_generic_filter:
+        assert field_name is not None
+        assert filter_value is not None
+        assert filter_type is not None
+        assert operator is not None
+        if field_name not in ChangeInfo.model_fields:
+            raise HTTPException(
+                status_code=422,
+                detail=f"Unknown changelog field: {field_name}.",
+            )
```
GibranAlfa: after some thoughts, this validation should happen in fmu-settings instead of here. I made this issue https://github.com/equinor/fmu-settings/issues/282 to improve the error handling of the filtering method. If that issue is resolved, then we don't need the validation here, and this can be removed. API will rely on its existing `ValueError` and convert it to 422 as it's done now in the endpoint.

So I suggest to remove this check and fix fmu-settings later.

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -384,10 +384,17 @@
     ),
     **inline_add_response(
         422,
-        "Invalid changelog data",
+        "Invalid changelog data or query parameters",
         [
             {"detail": "Invalid changelog format or data at {path}: {error}"},
             {"detail": "Invalid or corrupt JSON at {path}: {error}"},
+            {"detail": "Unknown changelog field: {field_name}."},
```
GibranAlfa: if the validation check is removed then this should also be removed

src/fmu_settings_api/deps/changelog.py
GibranAlfa: Annotated should not be added back

GibranAlfa (approved):
very nice

PR #419, opened 2026-06-29
CI: Speed up exception logging in tests

### LINKED ISSUE #417: FastAPI 0.137+ traceback locals make Rich console exception rendering slow
After scheduled CI picked up `fastapi>=0.137.0`, the test suite became much slower. The route tests with generic exception paths were especially affected.

Observed locally on `tests/test_v1/test_project.py`:

- After upgrade: ~403s
- Before upgrade: ~4-5s

The slow cases were tests such as `*_general_exception`, `*_unexpected_error`, and `*_other_exception`.

### Root Cause

FastAPI 0.137.0 changed how routers are wired internally. Instead of cloning routes, it now keeps the `APIRouter` / `APIRoute` structure as a tree. That means when an exception bubbles up through included routers, the traceback carries a lot more stuff.

Our dev console logging was using structlog’s default `ConsoleRenderer`, which uses Rich tracebacks with local variables enabled. Rich then tried to pretty-print all those huge locals for every exception-path test, and that made the tests slow.

### Proposed Solution

Console logging should uses plain traceback rendering instead of Rich exception rendering:

```python
structlog.dev.ConsoleRenderer(
    colors=True,
    exception_formatter=structlog.dev.plain_traceback,
)
```

GibranAlfa:
Resolves #417

Switch console exception logging to plain tracebacks so FastAPI 0.137+ router tracebacks do not make Rich render huge local variables. Tests now configure logging through the normal `setup_logging()` path with a strict mocked log manager, so direct `app` imports in pytest use the same console traceback behavior.

This keeps regular structlog console logs unchanged, but stops Rich from trying to render huge traceback locals for exception logs, which was what made the exception tests so slow in CI.

CI time:
Before this PR: 28m 54s (https://github.com/equinor/fmu-settings-api/actions/runs/28376766579)
After this PR: 42s (https://github.com/equinor/fmu-settings-api/actions/runs/28385731444)

## Checklist

### DIFF
--- a/src/fmu_settings_api/logging.py
+++ b/src/fmu_settings_api/logging.py
@@ -90,7 +90,10 @@ def setup_logging(
         ]
     else:
         processors += [
-            structlog.dev.ConsoleRenderer(colors=True),
+            structlog.dev.ConsoleRenderer(
+                colors=True,
+                exception_formatter=structlog.dev.plain_traceback,
+            ),
         ]
 
     structlog.configure(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,7 +6,7 @@ from collections.abc import AsyncGenerator, Callable, Generator, Iterator
 from contextlib import AbstractContextManager, contextmanager
 from pathlib import Path
 from typing import Any, Literal
-from unittest.mock import patch
+from unittest.mock import Mock, patch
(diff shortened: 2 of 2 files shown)


slangeveld (approved):
Wow, that's a major difference in CI run time. Nice!

PR #420, opened 2026-06-29
ENH: Add SMDA masterdata validation endpoint

### LINKED ISSUE #117: QC/Validate existing metadata with SMDA
We need a route that ensures the data configured in config.json is valid compared to the data in SMDA.

This should perhaps be a separate route paths, `lint/` or `validate/`, as other forms or more granular linting may be possible in the future. Perhaps on a resource basis.

But, some fields have extra dependencies, such as the `rms:` field which requires accessing the RMS project from an opened RMS instance via rmsapi.

GibranAlfa:
Resolves #117 

200 line counts from updating uv.lock and most lines are from tests.

- Added `POST /project/validate/masterdata/smda` to validate saved project SMDA masterdata against current SMDA data.
- Added `ProjectValidationService` and its deps. This service will also handle validation of RMS project later.
- Return structured mismatch details when saved masterdata no longer matches SMDA.
- Update `validation.masterdata_smda` metadata after successful validation.
- GUI needs to be updated later to parse the 422 error correctly, maybe a proper dialog showing what field that are mismatched, not just a toast

## Checklist


src/fmu_settings_api/v1/routes/project.py
```diff
@@ -236,6 +243,59 @@
...
+                            "message": (
+                                "Project config masterdata 'field' is not present "
+                                "in current SMDA data"
+                            ),
+                        }
+                    ],
+                }
+            },
+        ],
+    ),
```
GibranAlfa, line 283: example response of mismatch error

src/fmu_settings_api/services/project_validation.py
```diff
@@ -0,0 +1,145 @@
...
+            coordinate_systems=[
+                coordinate_system
+                for result in per_field_smda_results
+                for coordinate_system in result.coordinate_systems
+            ],
+        )
+
+        mismatches: list[ValidationMismatch] = []
+        for smda_attr in Smda.model_fields:
+            # SmdaMasterdataResult uses plural list fields for some saved
```
slangeveld: Took me quite some time to understand what was going on here, so this comment can probably be updated a bit to make it even more clear what is going on.

Maybe something like

```
SmdaMasterdataResult uses plural form for some of the attribute names compared to the attribute names in the Smda class used to save the SMDA data in .fmu. Therefore, we need to add an "s" when comparing some of the fields, e.g. coordinate_system must be compared against
coordinate_systems, and stratigraphic_column against stratigraphic_columns
```

src/fmu_settings_api/services/project_validation.py
```diff
@@ -0,0 +1,145 @@
...
+                        f"Project config masterdata '{smda_attr}' "
+                        "is not present in current SMDA data"
+                    ),
+                )
+            )
+
+        if mismatches:
+            raise MasterdataSmdaMismatchError(mismatches)
+
+        record = ValidationRecord(
```
slangeveld, line 141: Now that we do this more than one place (will need to do this other places as well with the RMS validation), we should maybe have a method "update_validation_metadata()" or similar in `fmu_settings` that creates the record and updates the metadata field in the config under the hood. Wdyt?

src/fmu_settings_api/v1/routes/project.py
```diff
@@ -851,6 +911,62 @@ async def patch_masterdata(
...
+        **GetSessionResponses,
+        **ProjectResponses,
+        **LockConflictResponses,
+        **MasterdataSmdaValidationResponses,
+    },
+)
+async def post_validate_masterdata_smda(
+    validation_service: ProjectValidationServiceDep,
+) -> Message:
+    """Validates saved project masterdata against SMDA."""
```
slangeveld: Here it says "saved project masterdata" while in the description above it is "saved project SMDA masterdata". I guess both are fine, but we should pick one and use that both places.

slangeveld (approved):
Looks good overall. Some comments/suggestions

src/fmu_settings_api/services/project_validation.py
GibranAlfa: I'm leaning towards keeping this in the API for now, because `fmu-settings` already has `set_config_value()` method that can update the config, the API should decide what to fill in the config. We can make a helper in `ProjectValidationService` to update the validation metadata, as RMS validation will also be done here and can use that helper.

PR #426, opened 2026-07-06
ENH: Add an endpoint to fetch well headers from SMDA

GibranAlfa:
Resolves #342

- Added `POST /api/v1/smda/well_headers`.
- Added `SmdaAPI.well_headers()` using `smda-api/wellheader-insight-headers/search`.
- Added `SmdaWellHeader` / `SmdaWellHeadersResult` models. Added some additional attributes in `SmdaWellHeader` that can be useful to identify the well. Documentation in docstrings are taken from SMDA directly.
- Added service logic to fetch well headers by field identifier.
- Added Drogon fallback data from `RMS_WELLS`.

## Checklist

### DIFF
--- a/tests/test_interfaces/test_smda_api.py
+++ b/tests/test_interfaces/test_smda_api.py
@@ -299,7 +299,26 @@ async def test_smda_discovery_search(mock_httpx_post: MagicMock) -> None:
 
 
 async def test_smda_well_header_search(mock_httpx_post: MagicMock) -> None:
-    """Tests well header search sends correct payload."""
+    """Tests well header search sends default payload."""
+    api = SmdaAPI("token", "key")
+
+    res = await api.well_headers(["FIELD_A"])
+
+    mock_httpx_post.assert_called_with(
+        f"{SmdaRoutes.BASE_URL}/{SmdaRoutes.WELL_HEADER_SEARCH}",
+        headers=api._headers,
+        json={
+            "_projection": "identifier,uuid",
+            "field_identifier": ["FIELD_A"],
+        },
+    )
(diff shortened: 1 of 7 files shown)


tests/test_interfaces/test_smda_api.py
```diff
@@ -296,3 +296,20 @@ async def test_smda_discovery_search(mock_httpx_post: MagicMock) -> None:
         },
     )
     res.raise_for_status.assert_called_once()  # type: ignore
+
+
+async def test_smda_well_header_search(mock_httpx_post: MagicMock) -> None:
```
slangeveld, line 301: The well_headers() methods contains an if statement so the tests here should cover both cases. Now it is only testing one

tests/test_v1/test_smda.py
```diff
@@ -1258,3 +1262,233 @@ async def test_post_strat_units_http_error(
...
     )
     assert "SMDA error requesting" in response.json()["detail"]
+
+
+async def test_post_well_headers_success(
+    client_with_smda_session: TestClient,
+    session_tmp_path: Path,
+    mock_SmdaAPI_post: AsyncMock,
+) -> None:
+    """Tests successful post to well_headers with valid field identifier."""
```
slangeveld, line 1270: We should not test the internals of the smda_service here as it adds unnecessary dependencies between the route and the service layer. It should be enough to test this logic in the service layer tests and in this test only verify that the service method is called when the endpoint is invoked.

Hence the test below where we check that drogon data is returned and we do not call SMDA is also unnecessary and should be covered in the smda_service tests.

slangeveld (approved):
Well (😄) done! A few comments on the tests

PR #432, opened 2026-07-13
MAINT: Better SMDA wellbore name for Drogon wells

### LINKED ISSUE #431: Change drogon well headers data to be similar to SMDA well names
Currently the Drogon well headers data is set to be the same as RMS well names, e.g. NO 55_33-A-1. However, its SMDA names should be "NO 55/33-A-1". We should change this to be similar to the real SMDA names.

GibranAlfa:
Resolves #431

Changed the SMDA wellbore names for Drogon wells to be more like SMDA names, from "NO 55_33-A-1" to "NO 55/33-A-1"

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/smda.py
+++ b/src/fmu_settings_api/services/smda.py
@@ -38,8 +38,8 @@ from fmu_settings_api.models.smda import (
 logger = get_logger(__name__)
 
 
-def _smda_wellbore_name(rms_well_name: str) -> str:
-    """Convert an RMS well name to an SMDA-style wellbore name."""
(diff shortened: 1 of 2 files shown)


src/fmu_settings_api/services/smda.py
```diff
@@ -36,6 +37,13 @@
 
 logger = get_logger(__name__)
 
+
+def _smda_wellbore_name(rms_well_name: str) -> str:
```
slangeveld: I guess this conversion only will make sense for Drogon data, so maybe we should name this method something with `Drogon`? The docstring should be updated accordingly

slangeveld (approved):

PR #448, opened 2026-08-11
ENH: Add azure logging support

### LINKED ISSUE #449: Add azure logging support
Currently the API only logs the user session to local user .fmu. We can upload this log to Azure to have centralized telemetry and observe how the application is used. Integration with `fmu-site-configuration` and `fmu-logging` package via `fmu-settings` would make this set up easier.

GibranAlfa:
Resolves #449 

Added optional Azure logging to `fmu-settings-api`. Structured API events at `INFO` and above are copied to Azure, while existing console and user-session logging keep their current behavior (defaults to logging CRITICAL event only if the user run 'fmu settings' command). This means that Azure log will always have log with INFO and above level, regardless of what the user set for log level via `--log-level` option in CLI. I think we want this behavior at first to see how the application is used, later we can easily change the log level for Azure if we want to.

Related PR to `fmu-site-configuration` https://github.com/equinor/fmu-site-configuration/pull/231.

## Flow

1. `fmu-settings-api` looks for a handler factory in the `fmu_settings_api` entry-point group.
2. `fmu-site-configuration` registers that entry point and creates the handler with the Azure connection string.
3. `fmu-logging` provides `FMUAzureLogHandler`, which sends the events through OpenTelemetry to Application Insights and Log Analytics.
4. The API flushes and closes the handler during shutdown.

Ownership: `fmu-settings-api` owns the events, fields, and log level, `fmu-site-configuration` provides the environment setup, and `fmu-logging` handles Azure transport. 

If the plugin is not installed or Azure setup fails, the API continues with local logging.

We can then query the log to see interesting metrics in a dashboard. Atlas group currently only has reader role in FMU-logging resource group, so I can't create a shared dashboard. We can ask the owner later to assign us contributor role, or ask them to make my private dashboard to a shared dashboard. Example of a dashboard:

image

One useful metric would be which assets use FMU Settings. The current logs do not contain this information, but we can add it later, for example by logging `Access.asset` for requests to `PATCH /project/access`.

## Checklist


src/fmu_settings_api/__main__.py
```diff
@@ -91,11 +91,11 @@ async def logging_request_validation_exception_handler(
 
 
 @asynccontextmanager
-async def lifespan(app: FastAPI) -> AsyncIterator[None]:
+async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
```
GibranAlfa, line 94: Small maintenance to change the return type to `AsyncGenerator`, which is more accurate because the function uses `yield`. It also resolves Pylance’s deprecation warning about using `AsyncIterator` with `@asynccontextmanager`. (ref [here](https://github.com/python/typeshed/pull/12087) about the deprecation warning)

src/fmu_settings_api/__main__.py
```diff
@@ -179,7 +184,13 @@ def run_server(  # noqa: PLR0913
     log_manager = UserSessionLogManager(user_fmu_dir)
 
     settings.log_level = log_level.upper()  # type: ignore[assignment]
-    setup_logging(settings, fmu_log_manager=log_manager, log_entry_class=EventInfo)
+    app.state.telemetry = setup_logging(
```
slangeveld: Is it possible to do this in two steps? First we set up logging, keeping the `setup_logging()` method unchanged. Then we have a new method `setup_telemetry()` that will be run if `enable_telemetry` is True. Will make it a bit easier to understand what is happening from reading the code here.

It is also good to separate concern as much as possible, keeping the logging code and telemetry code separate as the telemetry part is an additional feature to the logging itself and not part of the default flow.

GibranAlfa: good idea, implemented!

tests/test_main.py
```diff
@@ -72,20 +74,45 @@ async def get_value() -> dict[str, str]:
     assert asset_response.text == "app"
 
 
-def test_run_server_adds_frontend(tmp_path: Path) -> None:
-    """Add the frontend when its directory is supplied."""
+@pytest.mark.parametrize(
+    ("enable_telemetry", "run_id"),
+    [(False, None), (True, "run-123")],
+)
+def test_run_server_adds_frontend(
```
slangeveld: Maybe this should be split in two tests? Now it is testing two different cases at the same time which makes it hard to make a good method name and description of what it is testing. Might make more sense to split this into 
1. One test for testing that `run_server()` adds frontend when it should. This should also test that frontend is _not_ added when no frontend directory is set
2. One test for testing that `run_server()` adds telemetry when `enable_telemetry` is True (and not if False)

slangeveld (approved):
Looks good 👍 Nice and clean. Thanks for providing a good overview of the flow in the description

tests/test_main.py
GibranAlfa: Good point, I split them into two tests now.

PR #450, opened 2026-08-11
ENH: Add RMS project config validation

### LINKED ISSUE #443: Add an endpoint to update RMS validation in project config
A new method should be added in `ProjectValidationService`, that updates when RMS attribute in project config was last validated with the RMS project. The endpoint would then probably need an `RmsSession` opened, and put in the `/rms` route.

GibranAlfa:
Resolves #443 

Added `POST /api/v1/rms/validate` to compare saved RMS settings with the open RMS project.

The validation:

- compares the RMS version, horizons, zones, and wells
- ignores item order
- ignores the `planned` field when comparing wells
- returns all mismatches in a structured 422 response
- updates validation metadata only after successful validation
- preserves existing metadata and configuration after failed validation

The project RMS patch endpoints do not update RMS validation metadata because each endpoint saves only part of the RMS configuration. Unlike these endpoints, the masterdata patch saves the complete masterdata configuration, so it can update validation metadata.

The shared project validation service no longer requires SMDA access unless SMDA masterdata validation is requested.

The minimum `fmu-settings` version is updated to 1.1.0 to use the `update_validation_metadata()` method.

This issue https://github.com/equinor/fmu-settings-gui/issues/507 described some ideas on how to show the validation metadata and what happens when there are mismatches.

## Checklist

### DIFF
--- a/src/fmu_settings_api/services/project_validation.py
+++ b/src/fmu_settings_api/services/project_validation.py
@@ -1,17 +1,22 @@
 """Service for validating project configuration against external sources."""
 
 import asyncio
-import getpass
 import json
-from datetime import UTC, datetime
+from collections.abc import Sequence
 
 from fmu.datamodels.common import Smda
 from fmu.settings import ProjectFMUDirectory
-from fmu.settings.models.project_config import ValidationRecord
+from fmu.settings.models.project_config import (
+    RmsHorizon,
+    RmsStratigraphicZone,
+    RmsWell,
+)
 from pydantic import BaseModel
+from runrms.api import RmsApiProxy
 
 from fmu_settings_api.models.project import ValidationMismatch
 from fmu_settings_api.models.smda import SmdaMasterdataResult, SmdaSelectedField
+from fmu_settings_api.services.rms import RmsService
 from fmu_settings_api.services.smda import SmdaService
 
 
@@ -23,15 +28,25 @@ class MasterdataSmdaMismatchError(ValueError):
         self.mismatches = mismatches
 
 
+class RmsProjectMismatchError(ValueError):
+    """Raised when saved RMS settings do not match the open RMS project."""
+
+    def __init__(self, mismatches: list[ValidationMismatch]) -> None:
+        """Initialize with validation mismatches."""
+        self.mismatches = mismatches
+
+
 class ProjectValidationService:
     """Service for validating project configuration."""
 
-    def __init__(self, fmu_dir: ProjectFMUDirectory, smda_service: SmdaService) -> None:
-        """Initialize the service with project and SMDA access."""
+    def __init__(self, fmu_dir: ProjectFMUDirectory) -> None:
+        """Initialize the service with project access."""
         self._fmu_dir = fmu_dir
-        self._smda_service = smda_service
 
-    async def validate_masterdata_smda(self) -> None:
+    async def validate_masterdata_smda(
+        self,
+        smda_service: SmdaService,
+    ) -> None:
         """Validate saved SMDA masterdata and update validation metadata.
 
         Flow:
@@ -57,7 +72,7 @@ class ProjectValidationService:
         ]
         per_field_smda_results = await asyncio.gather(
             *[
-                self._smda_service.get_masterdata([selected_field])
+                smda_service.get_masterdata([selected_field])
                 for selected_field in selected_fields
             ]
         )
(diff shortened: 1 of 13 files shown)


src/fmu_settings_api/services/project_validation.py
```diff
@@ -57,7 +72,7 @@ async def validate_masterdata_smda(self) -> None:
         ]
         per_field_smda_results = await asyncio.gather(
```
slangeveld, line 73: Could we maybe do this smda lookup from the `post_validate_masterdata_smda` endpoint instead and just pass the result to this method? Then we don't need to pass the SmdaService at all. It is good practice to keep the responsibilities of the services as clear as possible and avoid having dependencies to other services.

This will make it easier to read and understand the code from the API layer and it will be easier to write unit tests for this validation service.

src/fmu_settings_api/services/project_validation.py
```diff
@@ -138,10 +153,134 @@ async def validate_masterdata_smda(self) -> None:
...
+        in the project configuration.
+        """
+        config = self._fmu_dir.config.load()
+        if config.rms is None:
+            raise ValueError("No RMS settings are saved in the FMU project.")
+
+        saved_rms = config.rms
+        mismatches: list[ValidationMismatch] = []
+
+        current_version = rms_service.get_rms_version(saved_rms.path)
```
slangeveld, line 186: Same as above: I think it would be good if we can fetch the relevant RMS data from the rms_service in the API layer and pass that data to this validation method. 

You might end up having to fetch some more data then needed, but it will make the architecture and structuring more clean, separating concern even more and avoid hidden dependencies.

src/fmu_settings_api/services/project_validation.py
```diff
@@ -138,10 +153,134 @@ async def validate_masterdata_smda(self) -> None:
...
             raise MasterdataSmdaMismatchError(mismatches)
 
-        record = ValidationRecord(
-            last_validated_at=datetime.now(UTC),
-            last_validated_by=getpass.getuser(),
-        )
-        self._fmu_dir.set_config_value(
-            "validation.masterdata_smda", record.model_dump()
-        )
+        self._fmu_dir.update_validation_metadata("masterdata_smda")
```
slangeveld, line 156: I think it would be good to update this validation metadata from the API layer. The validation service should only validate, not write anything to the resources.

src/fmu_settings_api/services/project_validation.py
```diff
@@ -138,10 +153,134 @@ async def validate_masterdata_smda(self) -> None:
...
+                    saved_items=saved_rms.wells,
+                    source_items=rms_service.get_wells(opened_rms_project),
+                    ignored_fields={"planned"},
+                )
+            )
+
+        if mismatches:
+            raise RmsProjectMismatchError(mismatches)
+
+        self._fmu_dir.update_validation_metadata("rms_project")
```
slangeveld, line 234: Same as above, would be good to do the updating of the validation metadata from the API layer and not in the validation service.

slangeveld:
Are we sure we dont want to update the RMS validation metadata in the project RMS patch endpoints? Even though it only updates parts of the RMS config, it is still an update that could potentially create a mismatch which is something it seems like we would like to validate. Not sure what is the best here, but maybe we should discuss this a bit more.

src/fmu_settings_api/services/project_validation.py
GibranAlfa: I would prefer to keep the SMDA lookup in the validation service. Fetching and combining the current SMDA data is part of the validation workflow, while the route should handle the request, call the service, and map errors. The service dependency is passed explicitly and can be mocked in the service tests.

src/fmu_settings_api/services/project_validation.py
GibranAlfa: I would prefer to keep the RMS lookups in the validation service for the same reason. The service decides which RMS values to fetch based on the saved configuration. Moving that orchestration into the route would give the route application logic and make the route tests depend on validation details. The RMS comparison logic is already separated in `_find_rms_item_mismatches()`.

src/fmu_settings_api/services/project_validation.py
GibranAlfa: I see `ProjectValidationService` as owning the complete validation operation: loading the saved configuration, fetching the current source data, comparing them, and updating validation metadata after success. Moving the metadata update to the route would make the route coordinate part of that workflow, so I would prefer to keep it here. I have renamed the method and updated its docstring to make this side effect explicit.

src/fmu_settings_api/services/project_validation.py
GibranAlfa: Same reasoning here. The service owns the complete validation operation, including updating validation metadata only after the full comparison succeeds. The route should call that operation and map its exceptions to HTTP responses, so I would prefer to keep the metadata update here. I have renamed the method and updated its docstring to make this side effect explicit.

GibranAlfa:
> Are we sure we dont want to update the RMS validation metadata in the project RMS patch endpoints? Even though it only updates parts of the RMS config, it is still an update that could potentially create a mismatch which is something it seems like we would like to validate. Not sure what is the best here, but maybe we should discuss this a bit more.

Good point. I have updated the RMS project path/version, stratigraphy, and wells update methods to clear the existing RMS validation metadata. The PATCH endpoints do not set a new validation timestamp because they do not validate the complete RMS configuration. Only a successful call to the RMS validation endpoint creates a new record. Coordinate-system updates are excluded because coordinate systems are not currently part of RMS validation.

src/fmu_settings_api/services/project_validation.py
slangeveld: Fair enough. It is just not so clear to me, when I read the method name `validate_masterdata_smda_and_update_metadata` in the API layer, that this this method will make a request to SMDA under the hood. Would be good to call it something like `validate_masterdata_against_smda_and_update_metadata`. But then you'll end up with a long name, which indicates to me that this method is doing too many things. So maybe we should split this in 2 methods, creating a wrapper in the validation service `update_validation_metadata` or similar that can be called by the API. From the API layer you can then have:
1. validation_service.validate_masterdata_against_smda(): performs the validation
2. validation_service.update_validation_metadata(): updates the validation status in the metadata

src/fmu_settings_api/services/project_validation.py
slangeveld: Yeah, I see the drawbacks of moving this out. If we can move the updating of the metadata out to a separate method `update_validation_metadata()` (or similar) as mentioned above, we can make it more clear from the method name that this service is validating agains the open RMS project through the RMS service. That should be clear enough I guess

src/fmu_settings_api/services/project_validation.py
GibranAlfa: In this case, I think it is better to keep the validation and metadata update in one method because they form one operation. If we split them, a caller could run the validation but forget to update the metadata. It would also make the route coordinate the validation workflow again. I renamed the method to `validate_masterdata_against_smda_and_update_metadata`

tests/test_services/test_project_validation_service.py
```diff
@@ -193,4 +233,282 @@ async def test_validate_masterdata_smda_raises_when_masterdata_is_missing(
...
+        await ProjectValidationService(
+            fmu_dir
+        ).validate_masterdata_against_smda_and_update_metadata(smda_service)
+
+
+def test_validate_rms_project_updates_validation_metadata(
+    fmu_dir: ProjectFMUDirectory,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Test matching RMS configuration writes a timezone-aware record."""
```
slangeveld: This docstring could be a bit clearer on what it is testing

tests/test_services/test_project_validation_service.py
```diff
@@ -193,4 +233,282 @@ async def test_validate_masterdata_smda_raises_when_masterdata_is_missing(
...
+    ]
+
+    ProjectValidationService(fmu_dir).validate_rms_project_and_update_metadata(
+        rms_service, Mock()
+    )
+
+    assert fmu_dir.config.load(force=True).validation.rms_project is not None
+
+
+def test_validate_rms_project_reports_all_mismatches(
```
slangeveld, line 366: Should there also be a test that verifies that no mismatches are returned when saved RMS items are equal to the RMS items in the open RMS project?

src/fmu_settings_api/services/project_validation.py
slangeveld: Good point. As long as the name is clear enough about what the method is doing (as it should be now), keeping like this is fine

slangeveld (approved):
Looks good :) A few more comments regarding tests to consider before merging

